PR #2, opened 2019-09-03
Wiring up form state and value conversion on submission

schottra:
* Reorganized all the form components into a subfolder.
* Created `ApiContext` to do dependency injection of api functions. This lets us mock API calls cleanly in stories/tests without having to hook into Axios and do ugly protobuf encoding of mock responses.
  * _Note_: I have only switched the hooks in use by this form and its stories to using the api context. Others will be migrated later.
* Implemented all the logic for selecting the workflow, fetching the launch plans, and using the default launch plan.
  * _Note_: I will add a launch plan selector in the next PR. For now, it's just using the default launch plan.
* Form now uses `WaitForData` where appropriate:
  * The initial form state must wait on fetching our 10 default workflow versions
  * The input rendering must wait on fetching of both workflows and launch plans. Otherwise it's going to render whatever input array we have set (which defaults to empty if no launch plan is selected)
* Added logic for converting the supported input values to their `Literal` versions for submission to the backend.
* Fixed z-index issue with workflow selector when other components are in the form with it
* Wired WorkflowSelector up to the form
* Fixed some behavior issues with datetime selector.

LaunchFormWithWFSelector


src/components/Launch/LaunchWorkflowForm/DatetimeInput.tsx
```diff
@@ -8,34 +9,33 @@ import { Moment, utc as moment } from 'moment';
...
-}
-
 /** A form field for selecting a date/time from a picker or entering it via
  * keyboard.
  */
 export const DatetimeInput: React.FC<InputProps> = props => {
     const { label, helperText, onChange, value: propValue } = props;
-    const value = typeof propValue === 'string' ? propValue : defaultDate();
+    const value =
+        typeof propValue === 'string' && propValue.length > 0
```
mrmcduff: Are you restricted from checking the type of this where `InputProps` is defined?

src/components/Launch/LaunchWorkflowForm/DatetimeInput.tsx
```diff
@@ -8,34 +9,33 @@ import { Moment, utc as moment } from 'moment';
...
     const handleChange = (
         dateValue: Moment | null,
         stringValue?: string | null
     ) => {
-        if (stringValue != null) {
-            onChange(stringValue);
-        } else if (dateValue !== null) {
+        if (dateValue && dateValue.isValid()) {
             onChange(dateValue.toISOString());
+        } else if (stringValue != null) {
```
mrmcduff: `!==` or are you deliberately coercing? If the latter, would you just want `else if (stringValue)`?

src/components/Launch/LaunchWorkflowForm/LaunchWorkflowForm.tsx
```diff
@@ -53,11 +58,38 @@ function getComponentForInput(input: InputProps) {
...
+                {
+                    key: 'version',
+                    operation: FilterOperationName.CONTAINS,
+                    value: query
+                }
+            ]
+        });
+        const options = workflowsToWorkflowSelectorOptions(workflows);
+        if (options.length > 0) {
+            options[0].description = 'latest';
```
mrmcduff: Is this something that is uniquely assigned here? Like, elsewhere that `workflowsToWorkflowSelectorOptions` is used, would you expect a nonzero length where the zeroth element is _not_ `latest`? Just checking to see if this is logic that should be contained in the function itself.

src/components/Launch/LaunchWorkflowForm/inputConverters.ts
```diff
@@ -0,0 +1,62 @@
+import { dateToTimestamp, millisecondsToDuration } from 'common/utils';
+import { Core } from 'flyteidl';
+import * as Long from 'long';
+import * as moment from 'moment';
```
mrmcduff: It doesn't look like this import is used. Also, I think you've capitalized `Moment` elsewhere when importing it in this fashion.

src/components/Launch/useDefaultLaunchPlan.ts
```diff
@@ -0,0 +1,36 @@
+import { useAPIContext } from 'components/data/apiContext';
+import { useFetchableData } from 'components/hooks';
+import { NotFoundError } from 'errors';
+import { LaunchPlan, WorkflowId } from 'models';
+
+/** Returns a fetchable for the default launch plan whose project/domain/name/version
+ * match the values in the given WorkflowId.
+ */
+export function useDefaultLaunchPlan(workflowId: WorkflowId | null = null) {
```
mrmcduff: Should this be organized under `hooks` somewhere? Or are you keeping ones that are intended only for specific components under that component.

src/components/Launch/LaunchWorkflowForm/DatetimeInput.tsx
schottra: Deliberately coercing. I want to allow empty string through, but not null/undefined.

schottra: Doh, although I do check above for empty string and change it to a date anyway...

src/components/Launch/LaunchWorkflowForm/LaunchWorkflowForm.tsx
schottra: This is specific to the form. I didn't want to hard-code that logic into the selector options function.

src/components/Launch/LaunchWorkflowForm/inputConverters.ts
schottra: It's used on line 28. My default behavior is to keep the import following the TSLint naming conventions (camelCased import). The one other place I do this, I'm importing and assigning `moment.utc` as `moment` so that all usages in the rest of the file are guaranteed to use the utc versions. I probably should do that here as well!

src/components/Launch/useDefaultLaunchPlan.ts
schottra: I'm trying to move away from keeping all the hooks in a common folder and instead colocating hooks in the modules that use them.

schottra:
Thanks @mrmcduff ! Since this is a feature branch, I'm going to merge and implement feedback in the next PR.

src/components/Launch/LaunchWorkflowForm/DatetimeInput.tsx
schottra: `InputProps.value` can be one of several types, so this is guarding against what happens if we pass a type we don't expect.
I have a planned cleanup pass in the future where I may just make all input values strings to simplify the logic a bit (It's also messy in the conversion code that runs on submit)

PR #3, opened 2019-09-03
Adds Launch Plan selector

schottra:
* Made `WorkflowSelector` a generic component, and made the fetchResults function optional. The default implementation will just search the values of the provided options without issuing any network requests
* Refactored the form state to make launch plans behave similarly to the workflow, instead of the weird mixture of selecting a launch plan or using a default one.
  * There is now an effect that fires once the launch plan loads which will attempt to select the default launch plan (the one with the same name as the workflow) if it exists.
* Used the new generic selector to allow selection of the launch plan
* Updated the logic for showing the inputs to wait until a launch plan is selected
LaunchPlanSelector


src/components/Launch/LaunchWorkflowForm/LaunchWorkflowForm.tsx
```diff
@@ -107,26 +108,42 @@ export const LaunchWorkflowForm: React.FC<LaunchWorkflowFormProps> = props => {
                     {...state.workflowOptionsLoadingState}
                 >
                     <div className={styles.formControl}>
-                        <WorkflowSelector
+                        <SearchableSelector
+                            label="Workflow Version"
```
mrmcduff: A const somewhere? `¯\_(ツ)_/¯ `

src/components/Launch/LaunchWorkflowForm/useLaunchWorkflowFormState.ts
```diff
@@ -240,28 +253,43 @@ export function useLaunchWorkflowFormState({
     useEffect(
         () => {
             const parsedInputs =
-                launchPlan && workflow.hasLoaded
-                    ? getInputs(workflow.value, launchPlan)
+                selectedLaunchPlan && workflow.hasLoaded
+                    ? getInputs(workflow.value, selectedLaunchPlan.data)
```
mrmcduff: Can `selectedLaunchPlan.data` also change, and should you trigger the `useEffect` on that too? Just checking.

src/components/Launch/LaunchWorkflowForm/LaunchWorkflowForm.tsx
schottra: It's only used in this file, in this single place. So I'm inclined not to make it a constant. If/when we get around to i18n support, my position on that will definitely change :-)

src/components/Launch/LaunchWorkflowForm/useLaunchWorkflowFormState.ts
schottra: Hmm. So technically, according to the Rules of Hooks, using that value means I should include it in the dependency array.

Practically, the data in the selector options is a const value that does not change unless you regenerate all of the items. In that case, the outer object would also change.

But I plan on enabling the extra lint rule in the future to ensure all captured values used in the effect are listed in the dependencies array. So I'll update it here.

Good catch!

schottra:
I'm going to merge this and make the above-mentioned change in my follow-up branch.

PR #8, opened 2019-10-11
Add support for project descriptions

schottra:
Project records now have optional description fields. This adds support for displaying the description in the cards shown in the project list. Projects with no description will show "(No description provided.)" as a placeholder.

* Added `ButtonLink` class to support using MUI link-style buttons with ReactRouter links under the hood
* Updated layout of cards in the projects list to look better with descriptions
* Updated the domain links to be Button-style links. This is more consistent with how MUI cards typically do actions

image

### DIFF
--- a/src/components/common/ButtonLink.tsx
+++ b/src/components/common/ButtonLink.tsx
@@ -1,0 +1,9 @@
+import * as React from 'react';
+import {
+    Link as RouterLink,
+    LinkProps as RouterLinkProps
+} from 'react-router-dom';
+
+export const ButtonLink = React.forwardRef<HTMLAnchorElement, RouterLinkProps>(
+    (props, ref) => <RouterLink innerRef={ref} {...props} />
+);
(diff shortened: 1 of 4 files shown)


src/components/common/ButtonLink.tsx
```diff
@@ -0,0 +1,9 @@
+import * as React from 'react';
+import {
+    Link as RouterLink,
+    LinkProps as RouterLinkProps
+} from 'react-router-dom';
+
+export const ButtonLink = React.forwardRef<HTMLAnchorElement, RouterLinkProps>(
+    (props, ref) => <RouterLink innerRef={ref} {...props} />
```
mrmcduff: QQ -- does the order have to be the same? In the generics declaration, you have `<HTMLAnchorElement, RouterLinkProps`, but then in the arguments, you have `(props, ref)`, where I would expect `(ref, props)` based on the declaration.

schottra: The order in the generic isn't necessarily related to the order of the arguments in the function. In the generic definition, the first argument is the type of the component (and thus the ref), and the second argument is the type of the props object.

PR #10, opened 2019-11-04
Exposes Workflow descriptions

schottra:
This builds on new additions to the Admin API to expose workflow descriptions. We previously had a placeholder for them in the Workflow Details page. But it was hard-coded with no value and always rendered a placeholder message. We are now actually fetching the description (using a new API), and we also show descriptions on the Workflow list page.

* Added API layer for fetching `NamedEntity` objects
* Added hooks for using `NamedEntity` objects
* Updated `SearchableWorkflowList` to use new hooks/types and show the workflow description
  * Also required a change to `SearchableList` to use `key` instead of `id`, as the latter is part of the `NamedEntity` type.
* Added a loading skeleton for fetching the workflow description to keep the page from jumping around when it is returned.

image

image


src/components/hooks/useNamedEntity.ts
```diff
@@ -0,0 +1,76 @@
...
+            defaultValue: {} as NamedEntity,
+            doFetch: id => getNamedEntity(id),
+            useCache: true
+        },
+        input
+    );
+}
+
+/** Fetches the NamedEntity record for a LaunchPlan */
+export function useLaunchPlanNamedEntity(
```
mrmcduff: Are these extra hooks intended to be in this PR? I don't see them used (yet)

schottra: I'm being proactive. They will be used in an upcoming feature.

PR #11, opened 2019-11-06
Tasks view

schottra:
This creates a view similar to the workflows list view, except for tasks.
For the moment, we just want to list the available tasks with their associated metadata (description). And a follow-up PR will also add the interface description for tasks.

* Added routes, side navigation for tasks list
* Updated `ProjectDetails` to support rendering different content based on the route (either workflows or tasks)
* Made most of `SearchableWorkflowNameList` generic so that it can be shared with the tasks list. The only difference between them is how the rows are rendered.
* Added hook for fetching task names/descriptions
* Added mock data for task ids/names

image

### DIFF
--- a/src/components/Task/SearchableTaskNameList.tsx
+++ b/src/components/Task/SearchableTaskNameList.tsx
@@ -1,15 +1,16 @@
 import { Typography } from '@material-ui/core';
 import ChevronRight from '@material-ui/icons/ChevronRight';
 import { SearchResult } from 'components/common';
-import { useCommonStyles } from 'components/common/styles';
 import {
     SearchableNamedEntity,
     SearchableNamedEntityList,
     SearchableNamedEntityListProps,
     useNamedEntityListStyles
-} from 'components/Workflow/SearchableNamedEntityList';
+} from 'components/common/SearchableNamedEntityList';
+import { useCommonStyles } from 'components/common/styles';
 import * as React from 'react';
 
+/** Renders a searchable list of Task names, with associated metadata */
 export const SearchableTaskNameList: React.FC<
     Omit<SearchableNamedEntityListProps, 'renderItem'>
 > = props => {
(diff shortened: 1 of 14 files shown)


src/components/Task/SearchableTaskNameList.tsx
```diff
@@ -0,0 +1,43 @@
...
+    SearchableNamedEntityList,
+    SearchableNamedEntityListProps,
+    useNamedEntityListStyles
+} from 'components/common/SearchableNamedEntityList';
+import { useCommonStyles } from 'components/common/styles';
+import * as React from 'react';
+
+/** Renders a searchable list of Task names, with associated metadata */
+export const SearchableTaskNameList: React.FC<
+    Omit<SearchableNamedEntityListProps, 'renderItem'>
```
mrmcduff: This is the first time I've learned about the `Omit` type. Neat.

src/components/Workflow/SearchableWorkflowNameList.tsx
```diff
@@ -1,135 +1,53 @@
...
-    </Typography>
-);
-
-interface SearchResultsProps {
-    results: SearchResult<SearchableWorkflowName>[];
-}
-const SearchResults: React.FC<SearchResultsProps> = ({ results }) => {
+/** Renders a searchable list of Workflow names, with associated descriptions */
+export const SearchableWorkflowNameList: React.FC<
+    Omit<SearchableNamedEntityListProps, 'renderItem'>
```
mrmcduff: If this type is common, it may be worth declaring someplace. Not a blocker, just noticing because I saw it twice.

schottra: Ah, yes. I'm making additional changes to this stuff in a follow-up branch. I'll move it to a common place if it stays common.

PR #19, opened 2019-11-08
Adding Task interface information to the list page

schottra:
This adds code to lazy-load and render task interface (inputs/outputs) information. It's using an IntersectionObserver to detect when the elements are about to be scrolled into view and firing an API call to retrieve the task definition for the associated row item.
If loading fails, we use a custom component for the error to render a less-obtrusive inline error string.

image

image

image


src/components/Task/SimpleTaskInterface.tsx
```diff
@@ -0,0 +1,86 @@
...
+const emptyVariables = {
+    variables: {}
+};
+
+const VariablesList: React.FC<{ variables: Record<string, Variable> }> = ({
+    variables
+}) => {
+    const commonStyles = useCommonStyles();
+    const styles = useStyles();
+    const output = Object.keys(variables).reduce<React.ReactNode[]>(
```
mrmcduff: Do you have a little test for this kind of reducer? I'm sure it works if you've got it running; for whatever reason the combination of the calls to `push` and returning the object feels counterintuitive to me.

schottra: I don't. But since this is code that is internal to the component, I think the right test would be ensuring that the component renders the correct strings for a few given combinations of values? I'll note it down and do a follow-up to add it later. Thank you!

PR #26, opened 2019-12-06
Integrate Authentication

schottra:
* Added hook for retrieving user information.
* Added api layer for fetching user information. By design, it will _never_ fail, but instead return null in the case of a 401. This is because Admin returns a 401 in both the case of not being logged in and the case of the user's session expiring.
* Centralized error handling logic for `/api/v1` requests.
* Updated `APIContext` to include information about the login state
* Updated `useFetchableData` to toggle the `expired` login state whenever we receive a 401.
* Added a global dialog component that listens for expired login state and shows a dialog directing the user to re-authenticate.
* Added error handling for 401, we now return a special error for that case (similar to 404).
* Updated `DataError` to render nothing in the case of a 401, since the assumption is that we will be showing a global dialog.
* Added AppBar component for displaying a login link or user name (if logged in)
* Removed usage of CORS proxy (Admin now fully supports CORS and proxying would not forward cookies)
* Updated `useFetchableData` to not return a rejected promise when errors occur. There are two reasons:
  * No code consuming this hook is calling `fetch` and chaining on the promise. It's designed to be waitable, but code should not be making decisions based on the return value of `fetch`. It should be using the state values exposed in the `FetchableData` returned by the hook.
  * Consuming code isn't expected to handle the rejection. This leads to errors about unhandled promise rejections. At the very least, it makes testing cumbersome when exercising failure cases.
* Updated testing libraries and added an override for micromatch to fix a performance issue with jest. Also changed order of module resolution in Jest to prefer `ts/tsx` first per [the recommendation of the docs](https://jestjs.io/docs/en/configuration#modulefileextensions-arraystring)
* Removed `snyk` for now. It's extremely slow during the installation process (slows down builds as well), and we haven't formally decided to use it.

image

image

### DIFF
--- a/src/components/data/apiContext.ts
+++ b/src/components/data/apiContext.ts
@@ -13,14 +13,25 @@ type APIFunctions = typeof CommonAPI &
     typeof TaskAPI &
     typeof WorkflowAPI;
 
-export interface APIContextValue extends APIFunctions {}
+export interface LoginStatus {
+    expired: boolean;
+    setExpired(expired: boolean): void;
+}
+export interface APIContextValue extends APIFunctions {
+    loginStatus: LoginStatus;
+}
+
 export const defaultAPIContextValue = {
     ...CommonAPI,
     ...ExecutionAPI,
     ...LaunchAPI,
     ...ProjectAPI,
     ...TaskAPI,
-    ...WorkflowAPI
+    ...WorkflowAPI,
+    loginStatus: {
+        expired: false,
+        setExpired: () => {}
+    }
 };
 
 /** Exposes all of the model layer api functions for use by data fetching
@@ -30,6 +41,21 @@ export const defaultAPIContextValue = {
 export const APIContext = React.createContext<APIContextValue>(
     defaultAPIContextValue
 );
+
+export function useLoginStatus(): LoginStatus {
+    const [expired, setExpired] = React.useState(false);
+    return {
+        expired,
+        setExpired
+    };
+}
+
+export function useAPIState(): APIContextValue {
+    return {
+        ...defaultAPIContextValue,
+        loginStatus: useLoginStatus()
+    };
+}
 
 export function useAPIContext() {
     return React.useContext(APIContext);
--- a/src/components/data/apiContext.ts
+++ b/src/components/data/apiContext.ts
@@ -42,7 +42,7 @@ export const APIContext = React.createContext<APIContextValue>(
     defaultAPIContextValue
 );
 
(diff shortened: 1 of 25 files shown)


src/components/data/apiContext.ts
```diff
@@ -13,14 +13,25 @@ type APIFunctions = typeof CommonAPI &
...
+
 export const defaultAPIContextValue = {
     ...CommonAPI,
     ...ExecutionAPI,
     ...LaunchAPI,
     ...ProjectAPI,
     ...TaskAPI,
-    ...WorkflowAPI
+    ...WorkflowAPI,
+    loginStatus: {
```
mrmcduff: Is this going to be incorporated into another API in the future? Feels somewhat odd to be the only item in the context not spread from some other imported value. Just a style thing, though.

src/components/hooks/useUserProfile.ts
```diff
@@ -0,0 +1,14 @@
+import { useAPIContext } from 'components/data/apiContext';
+import { useFetchableData } from 'components/hooks';
+import { UserProfile } from 'models';
+
+/** State hook that returns the user information if logged in, null otherwise */
+export function useUserProfile() {
+    const { getUserProfile } = useAPIContext();
```
mrmcduff: Does it make this hook somewhat large to be including the values in `useAPIContext`, which had a bunch of spread API objects? Just wondering if not atomizing it has any effect on perf, etc. (I have no evidence to support that it does; just realized that this one function was calling another that has larger scope)

src/components/data/apiContext.ts
schottra: I probably should have documented this!
This won't be part of an API. This is an application state value that tells us if a request has failed with a 401. It's part of this context for two reasons:
1. Any component/hook making an API request can see and/or set the loginStatus. At the moment, this is only used through the `useFetchableData` base hook.
2. APIContext gets exposed by a provider at the root of the component tree, allowing the blanket "your session expired" component to consume it.

src/components/hooks/useUserProfile.ts
schottra: Good question! This hook is consuming the value and not returning it. So I'm not concerned about the size. In most cases, it should be consuming the same value that it saw last time.

The broader question is still relevant: the API Context is now constructed by spreading the API functions and then adding the `loginStatus`. Until now, we were setting its default value once and then the Provider would just continue returning that.

Now, the value will be recreated each time the `App` component re-renders. However, that shouldn't be very often, because the only state it depends on that can change is the `loginStatus`.

PR #28, opened 2019-12-12
Ensure TaskExecutions are always listed in the correct order.

schottra:
See lyft/flyte#86

We have two places where we show a list of task executions. One of them was already doing sorting of the list by creation time, but was doing it in the component itself. Task executions support sorting when fetching from the API, so this should at least be done when requesting the data.
Since we always list them in order of creation, I put the sorting into the hook that is shared by both components.

* Removed explicit sort in `TaskExecutionsList`
* Updated `useTaskExecutions` to always request a sort by created_at, ascending
* Added tests for `TaskExecutionsList` and `NodeExecutionChildren` to ensure empty states and that the request includes the expected sort values
* Cleanup: Moved a string that appears as a literal in a few places into a constant

### DIFF
--- a/src/components/Executions/useTaskExecutions.ts
+++ b/src/components/Executions/useTaskExecutions.ts
@@ -1,13 +1,14 @@
+import { useAPIContext } from 'components/data/apiContext';
 import { every } from 'lodash';
 import {
     ExecutionData,
-    getTaskExecutionData,
     limits,
-    listTaskExecutions,
     NodeExecution,
     NodeExecutionIdentifier,
+    SortDirection,
     TaskExecution,
-    TaskExecutionIdentifier
+    TaskExecutionIdentifier,
+    taskSortFields
 } from 'models';
 import { useDataRefresher } from '../hooks';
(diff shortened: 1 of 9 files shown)


src/components/Executions/useTaskExecutions.ts
```diff
@@ -1,38 +1,42 @@
...
+    const { listTaskExecutions } = useAPIContext();
     return useFetchableData<TaskExecution[], NodeExecutionIdentifier>(
         {
             debugName: 'TaskExecutions',
             defaultValue: [],
-            doFetch: doFetchTaskExecutions
+            doFetch: async (id: NodeExecutionIdentifier) => {
+                const { entities } = await listTaskExecutions(id, {
+                    limit: limits.NONE,
+                    sort: {
```
mrmcduff: Does this replace the sorting that was removed in the earlier file?

schottra: It does. The earlier file was doing the sorting client-side. This uses the sorting built in to the API so that the results come back already in the desired order.

PR #30, opened 2019-12-18
Add support for Collections in the launch UI

schottra:
lyft/flyte#22

(This is being opened against a feature branch. I would like to add some more handling of errors and mapping to the fields in the form before merging this to master)

[Collections](https://github.com/lyft/flyteidl/blob/master/protos/flyteidl/core/literals.proto#L83) are a list of Literal values which all have the same type. The inner type of a Collection may be another Collection, making nesting possible.

Our initial support for collections will require users to enter values as a valid string of JSON, which we will parse into native JS(ON) types (string, number, boolean) and then convert to the appropriate `Literal` type based on the type of the collection. It's a little clunky, but easy to validate and uses a well-known standard.

Some important points:
* We're using `lossless-json` for the parsing because it preserves integers that would otherwise overflow and gives us the ability to directly convert them to `Long`s (which are used in our protobuf message classes to handle int64s).
* For boolean, we will allow a handful of truthy/falsey values for convenience (`0, 1, '0', '1', true, false, 'true', 'false', 't', 'f'`)
* Because it's a JSON string, the user must add their own brackets. We _could_ try and do this for them, but since collections can contain nested collections, trying to handle the edge cases of how many brackets to add doesn't seem worth the complication.
* Also because it's JSON, string values must be in double quotes (`""`). This differs from entering single string values, which do not require quotes.

I also made some changes to the input conversion functions:
* The input now contains the type information (collections need this to determine the nested converter to call).
* The input value is now an `InputValue` instead of being explicitly `toString`-ed beforehand. This puts the responsibility on the conversion function for converting to a string if necessary
* Most of the input converters can now accept the native type in addition to a string value. This was needed to support the collection parsing case.

And I also added a _lot_ of test cases to cover all of the expected variations for values to be parsed, collections of those value types, and singly-nested collections of those value types.


src/components/Launch/LaunchWorkflowForm/CollectionInput.tsx
```diff
@@ -0,0 +1,55 @@
...
+        case InputType.Map:
+        case InputType.String:
+        case InputType.Struct:
+            return (
+                <TextField
+                    helperText={helperText}
+                    fullWidth={true}
+                    label={label}
+                    multiline={true}
+                    onChange={stringChangeHandler(onChange)}
```
mrmcduff: Where does the input value to `stringChangeHandler` here get defined? Is it recursively referencing the `onChange` handler of the `TextField`? So, `x = stringChangeHandler(x)`?

src/components/Launch/LaunchWorkflowForm/inputConverters.ts
```diff
@@ -1,62 +1,166 @@
...
+    InputProps,
+    InputType,
+    InputTypeDefinition,
+    InputValue
+} from './types';
 
-function booleanToLiteral(value: string): Core.ILiteral {
-    return { scalar: { primitive: { boolean: Boolean(value) } } };
+function losslessReviver(key: string, value: any) {
+    if (value && value.isLosslessNumber) {
```
mrmcduff: Does this handle a value of `0`?

schottra: I see why you're asking (0 is falsey). Your intuition is correct, but it should be fine in this case, as the behavior is just to return `value` directly. That would result in a `number` and is the desired behavior.

src/components/Launch/LaunchWorkflowForm/CollectionInput.tsx
schottra: `stringChangeHandler` generates an event handler that will forward the value received to its argument. So in this case, the prop value ends up looking like:
```
onChange = ({ target: { value } }) => props.onChange(value); // props here are the props to CollectionInput
```

PR #32, opened 2020-01-06
Launch tests

schottra:
* Added tests for the form behavior in various cases
* Updated the way some of the components render to make them easier to find with the various query function from RTL
* Updated validation logic for the various simple types
* Filled out test cases for the validation logic


src/components/Launch/LaunchWorkflowForm/DatetimeInput.tsx
```diff
@@ -7,12 +7,20 @@ import {
...
 import * as React from 'react';
 import { InputProps } from './types';
+import { getLaunchInputId } from './utils';
+
+/* This module imports differently in tests and causes errors.
+ * The long-term solution is to enable `esmModuleInterop` and change all of
+ * our `import * as` to be normal default imports. But as this is the only
+ * module currently exhibiting the problem, this is a less impactful change.
+ */
+const momentDateUtils = momentUtils.default ? momentUtils.default : momentUtils;
```
mrmcduff: Thanks for explaining why this is happening!

src/components/Launch/LaunchWorkflowForm/inputHelpers/datetime.ts
```diff
@@ -1,18 +1,29 @@
...
 import { Core } from 'flyteidl';
 import { utc as moment } from 'moment';
+import { InputValue } from '../types';
+import { allowedDateFormats } from './constants';
 import { ConverterInput, InputHelper } from './types';
 
+function parseDate(value: InputValue) {
+    return value instanceof Date
+        ? value
+        : moment(value.toString(), allowedDateFormats).toDate();
```
mrmcduff: I didn't know you could put in an array here -- does it pick one of them preferentially if there's ambiguity to the format?

schottra: I just learned about it: https://momentjs.com/guides/#/parsing/multiple-formats/
1. If the date isn't in one of the allowed formats, it results in an invalid date, instead of falling back to the JS `Date` constructor
2. It tries them in the order specified.

PR #37, opened 2020-02-04
Adds support for initializing the Launch form using a workflow/launch plan/ inputs

schottra:
lyft/flyte#84

This is the bulk of the work needed to support a "Relaunch with edits" experience. It adds the ability for the launch form state to be initialized from specific parameters.

**Workflow Version**
The launch form can be pre-initialized to point to a specific workflow version. The provided version will be included in the fetching of workflow versions to show in the selector, and we will attempt to select that version after loading the options. Fallback behavior is to select the first item in the list (the latest version)

**Launch Plan**
Providing a launch plan id will cause that launch plan to be preferred. Similar to workflow versions, it will be included in the fetching of options. If we can't select the desired launch plan, we fall back to the first item in the list.
_Additionally_, when changing workflow versions, we will now preserve the selected launch plan **name** and attempt to select it before falling back to the first item in the list. This allows the user to change versions without losing their selected launch plan

**Input values**
A `Map` of `InputValue`s can also be provided. The keys for the map are a computed cache key based on the `type` of the input as well as its `name`. This ensures that we don't attempt to use a provided value for an input with the same name but a different type (for example, changing the type of an input between one version of a workflow and another).

_Caching of Input Values_
Values entered into the form (whether loaded from an execution or entered by the user) will be pushed into a cache held in a Context provider. When changing versions/launch plans, we will re-initialize the inputs using their cached values if possible. This allows users to change versions or launch plans without losing the values they entered.

The change also includes unit tests for most of the above-mentioned behavior, and a Storybook story that exercises the "launch from execution" behavior.

image

(Preserving input values across version/launch plan changes)
PreserveInputsAfterFormChange

### DIFF
--- a/src/components/Launch/LaunchWorkflowForm/LaunchWorkflowForm.tsx
+++ b/src/components/Launch/LaunchWorkflowForm/LaunchWorkflowForm.tsx
@@ -17,6 +17,7 @@ import {
 } from 'models';
 import * as React from 'react';
 import { formStrings } from './constants';
+import { InputValueCacheContext } from './inputValueCache';
 import { LaunchWorkflowFormInputs } from './LaunchWorkflowFormInputs';
 import { SearchableSelector } from './SearchableSelector';
 import { useStyles } from './styles';
@@ -63,7 +64,7 @@ export const LaunchWorkflowForm: React.FC<LaunchWorkflowFormProps> = props => {
     };
 
     return (
-        <>
+        <InputValueCacheContext.Provider value={state.inputValueCache}>
             <DialogTitle disableTypography={true} className={styles.header}>
                 <div className={styles.inputLabel}>{formStrings.title}</div>
                 <Typography variant="h6">{state.workflowName}</Typography>
@@ -152,6 +153,6 @@ export const LaunchWorkflowForm: React.FC<LaunchWorkflowFormProps> = props => {
                     </Button>
                 </DialogActions>
             </div>
-        </>
+        </InputValueCacheContext.Provider>
     );
 };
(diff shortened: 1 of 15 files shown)


src/components/Launch/LaunchWorkflowForm/useLaunchWorkflowFormState.ts
```diff
@@ -213,32 +320,70 @@ export function useLaunchWorkflowFormState({
...
+                ({ name }) => name === lastSelectedLaunchPlanName
+            );
+            if (lastSelected) {
+                onSelectLaunchPlan(lastSelected);
+                return;
+            }
+        }
+
+        if (preferredLaunchPlanId) {
+            const preferred = launchPlanSelectorOptions.find(
```
mrmcduff: We're not worried about this being an intentional `''`, right?

src/components/Launch/LaunchWorkflowForm/LaunchWorkflowForm.tsx
```diff
@@ -63,7 +64,7 @@ export const LaunchWorkflowForm: React.FC<LaunchWorkflowFormProps> = props => {
     };
 
     return (
-        <>
+        <InputValueCacheContext.Provider value={state.inputValueCache}>
```
mrmcduff: This is a neat way of using hooks in place of state for an FC; I missed the purpose of that hook last time.

src/components/Launch/LaunchWorkflowForm/useLaunchWorkflowFormState.ts
schottra: The `id` property is an object, so no. It's either defined with the expected fields, or it's falsey and we want to ignore it.

PR #40, opened 2020-02-12
Updating dependencies to fix security vulnerability

schottra:
There is a vulnerability in `set-value`, which is a deep dependency via a few different packages we are using. Upgrading some of those packages naturally led to other things breaking. Most were fixed by type renames or increased specificity. The only real code change was updating some `act` statements to use a wrapping inner arrow function instead of calling `jest.runAllTimers` directly. This should be functionally equivalent though, if a tiny tiny bit less efficient due to generating a few extra functions in memory :grin:


src/components/flytegraph/Layout.tsx
```diff
@@ -29,7 +29,7 @@ export class Layout<T> extends React.Component<LayoutProps<T>> {
     // input props actually change
     layoutGraph = memoizeOne<typeof layoutGraph>(
         (data, config) => layoutGraph(data, config),
-        layoutArgsAreEqual
+        layoutArgsAreEqual as EqualityFn
```
BobNisco: Is there a legitimate issue that we're getting around by using `as` here?

schottra: The compiler was having trouble reconciling the type of `layoutArgsAreEqual` with the type `EqualityFn`. The latter accepts `any[]` as its input, the former accepts `[GraphInputNode[], GraphConfig]`. Those _should_ be compatible, but I get `Type 'any[]' is missing the following properties from type '[GraphInputNode[], GraphConfig]': 0, 1`. I'm giving it a little nudge to help.

src/components/Launch/LaunchWorkflowForm/test/LaunchWorkflowForm.test.tsx
```diff
@@ -287,7 +289,9 @@ describe('LaunchWorkflowForm', () => {
             expect(integerInput).toBeInvalid();
 
             fireEvent.change(integerInput, { target: { value: '123' } });
-            act(jest.runAllTimers);
+            act(() => {
+                jest.runAllTimers;
+            });
             await wait();
             expect(integerInput).toBeValid();
```
BobNisco: This assertion seems to be failing in CI

BobNisco: Along with the assertion on line 472 of this file

schottra: Yep, looking at those failures now 👀

PR #48, opened 2020-02-20
Fixes initial population of values for string Collections in the launch form.

schottra:
lyft/flyte#181

When converting `Literal` values to `InputValue`s for use in the form, the `Collection` type is output as a JSON representation of an array. The existing implementation was blindly wrapping `[]` around values and comma-separating them. For Dates and strings, the resulting array string is not valid JSON and will trigger validation errors.

It turned out to be easier to allow `JSON.stringify` to do this work for us, with the exception of integers. Since we use a special type for integers, `stringify` would result in a quoted string value instead of a bare integer value.

### DIFF
--- a/src/components/Launch/LaunchWorkflowForm/inputHelpers/utils.ts
+++ b/src/components/Launch/LaunchWorkflowForm/inputHelpers/utils.ts
@@ -23,10 +23,5 @@ export function collectionChildToString(type: InputType, value: any) {
     if (value === undefined) {
         return '';
     }
-    switch (type) {
-        case InputType.Integer:
-            return `${value}`;
-        default:
-            return JSON.stringify(value);
-    }
+    return type === InputType.Integer ? `${value}` : JSON.stringify(value);
 }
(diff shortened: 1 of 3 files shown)


src/components/Launch/LaunchWorkflowForm/inputHelpers/utils.ts
```diff
@@ -14,3 +15,13 @@ export function extractLiteralWithCheck<T>(
     }
     return value as T;
 }
+
+/** Converts a value within a collection to the appropriate string
+ * representation. Some values require additional quotes.
+ */
+export function collectionChildToString(type: InputType, value: any) {
```
BobNisco: Could be worthwhile to add an explicit return type of `string` here so that future code changes do not accidentally return unexpected types

schottra: That's a good idea. Code consuming this expects a string, so changes here would still result in a type error. But I can't guarantee that any new usage of this function would also have such a strict type expectation.

PR #49, opened 2020-02-21
Select contents of workflow version selector on focus

schottra:
lyft/flyte#140

Previously, we would clear the value of this input every time the user focused, so that they could begin typing right away. The goal was convenience. But it makes it difficult to copy the value from the input.

A solution that covers both needs is to simply highlight the contents on focus, so they are copyable and so that typing will have the effect of clearing the existing value.

DontClearWFVersionOnClick

### DIFF
--- a/src/components/Launch/LaunchWorkflowForm/SearchableSelector.tsx
+++ b/src/components/Launch/LaunchWorkflowForm/SearchableSelector.tsx
@@ -91,6 +91,7 @@ function useSearchableSelectorState<DataType>({
     onSelectionChanged
 }: SearchableSelectorProps<DataType>): SearchableSelectorState<DataType> {
     const fetchResults = fetchSearchResults || generateDefaultFetch(options);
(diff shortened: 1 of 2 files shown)


src/components/Launch/LaunchWorkflowForm/SearchableSelector.tsx
```diff
@@ -127,18 +129,21 @@ function useSearchableSelectorState<DataType>({
     };
 
     const onFocus = () => {
+        setIsExpanded(false);
+        setHasReceivedInput(false);
```
mrmcduff: Does order here matter? Or do the hooks complete asynchronously?

schottra: My understanding is that state updates triggered in response to an event are batched. So the order here won't matter. But, even if that turns out not to be the case, the order here is alright.

PR #50, opened 2020-02-25
Preserves unsupported input types when relaunching

schottra:
lyft/flyte#186

For some input types, we don't yet have editing support in Flyte Console. This caused a regression when changing our relaunch flow to use the form (see the above issue). 

However, it's possible for us to preserve the `Literal` that we extracted from the source execution and pass it along unmodified when we relaunch the execution. This change enables that behavior by changing the types passed down to our input state to be a `Core.ILiteral` until it is actually necessary to convert it to an `InputValue`. When the form is submitted, if there is no user-provided value, but there *is* an initial value, we will send that instead.

* Updated `ParsedInput` and `InputProps` to add the `initialValue` field. Removed the `defaultValue` field.
* Moved the logic for converting from `Literal` to `InputValue` down to the input state hook.
* Moved logic for using the default `InputValue` into the input state hook.
* Updated validation to only throw errors if a required input does not have an initial value
* Updated `inputToLiteral` to return the `initialValue` property if provided and the user entered no value.
* No longer setting any values into the InputCache from the form state hook. It's only used by the input state hook to preserve user-provided values. Default values and values initialized from an execution will be passed down via the `initialValue` properties and parsed by the input state hook.

### DIFF
--- a/src/components/Launch/LaunchWorkflowForm/useLaunchWorkflowFormState.ts
+++ b/src/components/Launch/LaunchWorkflowForm/useLaunchWorkflowFormState.ts
@@ -221,10 +221,7 @@ export function useLaunchWorkflowFormState({
     >();
     const selectedWorkflowId = selectedWorkflow ? selectedWorkflow.data : null;
 
-    const inputValueCache = useMemo(
-        () => createInputValueCache(initialParameters.values),
-        [initialParameters.values]
-    );
+    const [inputValueCache] = useState(createInputValueCache());
 
     // We have to do a single item get once a workflow is selected so that we
     // receive the full workflow spec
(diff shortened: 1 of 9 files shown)


src/components/Launch/LaunchWorkflowForm/useLaunchWorkflowFormState.ts
```diff
@@ -221,10 +221,7 @@ export function useLaunchWorkflowFormState({
     >();
     const selectedWorkflowId = selectedWorkflow ? selectedWorkflow.data : null;
 
-    const inputValueCache = useMemo(
-        () => createInputValueCache(initialParameters.values),
-        [initialParameters.values]
-    );
+    const [inputValueCache] = useState(createInputValueCache());
```
BobNisco: Were you running into issues with `useMemo`? I guess I'm struggling to see why we'd use `useState` here when the value doesn't seem to be mutated ever

schottra: Nope, no issues. `useMemo` is overly complicated here now. We don't insert any values into the cache anymore at this level. So we just want to create it exactly once and reuse the value for the life of the component.

PR #56, opened 2020-04-03
Status banner

schottra:
lyft/flyte#223

This implements the status banner that will show whatever status/message is retrieved from the `STATUS_URL` endpoint. There are two possible icons (info and warning). The banner will only show if the `message` field of the status is provided. The banner is dismissable per-session and will return on refresh for as long as the status message exists.

Still to do in future PRs:
* Tests
* Automatically parse and render links

image

image

image

### DIFF
--- a/src/components/Notifications/SystemStatusBanner.tsx
+++ b/src/components/Notifications/SystemStatusBanner.tsx
@@ -1,0 +1,59 @@
+import Paper from '@material-ui/core/Paper';
+import { makeStyles, Theme } from '@material-ui/core/styles';
+import { WaitForData } from 'components/common';
+import { SystemStatus } from 'models/Common/types';
+import * as React from 'react';
+import { useSystemStatus } from './useSystemStatus';
+
+const useStyles = makeStyles((theme: Theme) => ({
+    statusPaper: {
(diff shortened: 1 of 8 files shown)


src/components/Notifications/SystemStatusBanner.tsx
```diff
@@ -0,0 +1,136 @@
...
+export const SystemStatusBanner: React.FC<{}> = () => {
+    const systemStatus = useSystemStatus();
+    const [dismissed, setDismissed] = React.useState(false);
+    const onClose = () => setDismissed(true);
+    if (dismissed) {
+        return null;
+    }
+    return (
+        <WaitForData {...systemStatus}>
+            {systemStatus.value.message && (
```
BobNisco, line 128: Is there a chance that `systemStatus.value.message` will be `undefined`? If so, I'd suggest turning this into a ternary to explicitly render `null` since `undefined` will blow up React at runtime

schottra: Oh right, I usually do `!!` for this case. I'll update it to a ternary in the following PR. Thanks for catching it.

PR #57, opened 2020-04-06
Status banner cleanup

schottra:
lyft/flyte#223

This wraps up the work to display a status banner based on the presence of `env.STATUS_URL` and a valid message payload.

* Fixed some layout issues with the banner that caused the text to not fill the available space
* Added tests for existing code
* Brought in `linkify-it`, added utility function to split a string of text into chunks based on the presence of valid links.
* Added a `LinkifiedText` component which uses the new utility function to split a string and then renders any links found as inline anchor tags which open in a new tab

image


src/components/Notifications/test/SystemStatusBanner.test.tsx
```diff
@@ -0,0 +1,91 @@
...
+        );
+
+    it('should display an info icon for normal status', async () => {
+        const { getByTestId } = renderStatusBanner();
+        await wait();
+
+        expect(getByTestId('info-icon')).toBeInTheDocument();
+    });
+
+    it('should display a waring icon for degraded status', async () => {
```
podehnal, line 38: typo `waring`

src/components/Notifications/SystemStatusBanner.tsx
```diff
@@ -125,12 +136,12 @@ export const SystemStatusBanner: React.FC<{}> = () => {
     }
     return (
         <WaitForData {...systemStatus}>
-            {systemStatus.value.message && (
+            {systemStatus.value.message ? (
```
BobNisco, line 139: 😄  Thanks for the follow-up here!

src/components/Notifications/test/SystemStatusBanner.test.tsx
schottra: oops :-)

PR #59, opened 2020-04-07
Handle new TaskExecutionPhase values

schottra:
* Upgraded `flyteidl`
* Added entries to task execution phase constants to handle the two new phase values ("waiting for resources" and "initializing")
* Fixed a few other errors in mock data due to some proto messages getting additional required fields.

image

image

### DIFF
--- a/src/models/Execution/__mocks__/mockTaskExecutionsData.ts
+++ b/src/models/Execution/__mocks__/mockTaskExecutionsData.ts
@@ -1,6 +1,6 @@
 import { dateToTimestamp, millisecondsToDuration } from 'common/utils';
 import { Admin, Core } from 'flyteidl';
-import { cloneDeep, random } from 'lodash';
+import { cloneDeep } from 'lodash';
 import { NodeExecutionPhase, TaskExecutionPhase } from '../enums';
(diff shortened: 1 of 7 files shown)


src/models/Execution/__mocks__/mockTaskExecutionsData.ts
```diff
@@ -51,18 +51,15 @@ export const createMockTaskExecutionsListResponse = (length: number) => {
             const startedAt = dateToTimestamp(
                 new Date(Date.now() - 1000 * 60 * (idx + 1))
             );
-            const phase =
-                idx < length - 1
-                    ? TaskExecutionPhase.FAILED
-                    : TaskExecutionPhase.SUCCEEDED;
+            const phase = random(Object.keys(TaskExecutionPhase).length - 1);
```
BobNisco, line 54: I don’t think random is a good idea in test data due to it breaking the concept of determinism in tests. Can we rework this so it’s consistently reproducible?

schottra: I probably should have left a comment. These are only used in stories, so the randomness, is desirable because it provides variation to the data when viewing it.

PR #61, opened 2020-04-15
Code Coverage published to codecov.io

kumare3:
# TL;DR
The PR publishes test coverage report to codecov.io. This allows tracking every checkin and the code coverage over time

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
Same as TL;DR

## Tracking Issue
https://github.com/lyft/flyte/issues/260

## Follow-up issue
NA

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,6 +33,6 @@ clean:
 	yarn run clean
 
 .PHONY: test_unit_codecov
-test_unit_codecov:
+test_unit_codecov: test_unit
 	npm install codecov -g
 	codecov
--- a/Makefile
+++ b/Makefile
@@ -31,3 +31,8 @@ server:
 .PHONY: clean
 clean:
 	yarn run clean
+
+.PHONY: test_unit_codecov
+test_unit_codecov:
+	npm install codecov -g
+	codecov
--- a/Makefile
+++ b/Makefile
@@ -35,4 +35,4 @@ clean:
 .PHONY: test_unit_codecov
 test_unit_codecov: test_unit
 	npm install codecov -g
-	codecov
+	codecov -f .coverage/coverage-final.json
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -31,3 +31,8 @@ server:
 .PHONY: clean
 clean:
 	yarn run clean
+
+.PHONY: test_unit_codecov
+test_unit_codecov: test_unit
+	npm install codecov -g
```
schottra, line 37: This should just be added to the project dependencies, not installed every time we run make. Also, we use `yarn` for dependency management. Using `npm` here will probably generate a separate lock file that is incompatible with our yarn lockfile.

Makefile
```diff
@@ -31,3 +31,8 @@ server:
...
+	codecov -f .coverage/coverage-final.json
```
schottra, line 38: This doesn't appear to actually run the unit tests anymore, just generates code coverage. I think we need something like `yarn test && yarn run codecov`.

package.json
```diff
@@ -20,7 +20,7 @@
     "start": "node -r dotenv/config index.js",
     "storybook": "start-storybook -p 9001 -c .storybook",
     "tdd": "yarn run test --watch --verbose false",
-    "test": "NODE_ENV=test jest --config=jest.config.js"
+    "test": "NODE_ENV=test jest --coverage --config=jest.config.js"
```
schottra, line 23: Generating coverage is a lot slower than just running the tests. I would recommend we create a separate script here just to run code coverage, and not run it every time unit tests are run.

Makefile
schottra: Nvmd, I missed that it has a dependency on the other make target

PR #62, opened 2020-04-15
Removes global package install for codecov

schottra:
This removes the usage of npm and a global package install during our code coverage make target. It also separates the coverage out into a separate testing command, since generating coverage slows down the local development flow and only needs to be run in the CI environment.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ install: #installs dependencies
 
 .PHONY: install_ci
 install_ci: install
-    yarn add codecov
+	yarn add codecov
 
 .PHONY: lint
 lint: #lints the package for common code smells
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,8 @@ server:
 clean:
 	yarn run clean
 
+# test_unit_codecov runs unit tests with code coverage turned on and
+# submits the coverage to codecov.io
 .PHONY: test_unit_codecov
-test_unit_codecov: test_unit
-	npm install codecov -g
-	codecov -f .coverage/coverage-final.json
+test_unit_codecov:
+	yarn run test:coverage && yarn run codecov
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,10 @@ update_boilerplate:
 install: #installs dependencies
 	yarn
 
+.PHONY: install_ci
+install_ci: install
+    yarn add codecov
+
 .PHONY: lint
 lint: #lints the package for common code smells
 	yarn run lint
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -20,7 +21,8 @@
     "start": "node -r dotenv/config index.js",
     "storybook": "start-storybook -p 9001 -c .storybook",
     "tdd": "yarn run test --watch --verbose false",
-    "test": "NODE_ENV=test jest --coverage --config=jest.config.js"
+    "test": "NODE_ENV=test jest --config=jest.config.js",
+    "test:coverage": "NODE_ENV=test jest --coverage --config=jest.config.js"
```
kumare3, line 25: i actually dont know if --coverage is required anymore, can you just try dropping `test:coverage`

Makefile
```diff
@@ -32,7 +32,8 @@ server:
...
 	yarn run clean
 
+# test_unit_codecov runs unit tests with code coverage turned on and
+# submits the coverage to codecov.io
 .PHONY: test_unit_codecov
-test_unit_codecov: test_unit
-	npm install codecov -g
-	codecov -f .coverage/coverage-final.json
+test_unit_codecov:
+	yarn run test:coverage && yarn run codecov
```
kumare3, line 39: does yarn run automatically install codecov?

kumare3: ohh nevermind you added it as a dependency. But why?
Its not really required to install or execute the code. IMO this is not ideal

package.json
schottra: It's required. That flag causes jest to generate the coverage data, which the `codecov` tool will subsequently process and upload. We just don't want to generate that data every time unit tests are run.

Makefile
schottra: Having it as a global dependency (even in travis) is also not ideal. This way, the version we are using is stored in our lockfile and will remain consistent unless we specifically upgrade. It's also stored as a dev dependency, not a runtime dependency. 
I can move out of our dependencies and just install it during the travis builds. But either way, it should probably be installed with yarn, since that's what we use for package management.

schottra: @kumare3 I updated it to install codecov only on Travis as part of the `install` step. It's important to do it here because Travis caches the result of the `install` step to make future runs faster. I think that's the middle ground between what I was trying to do and your suggestion (keeps it out of our dependencies for installing/running, keeps it out of global package scope, allows it to be cached by Travis).

kumare3: ya sorry I am not really aware of the tooling, yesterday was the first time I was playing around with all the react and node tooling. I learnt a little bit - jest, npm, yarn etc

package.json
kumare3: hmm when i run npm test it always generates a coverage, especially because jest.config has coverage enabled?

PR #67, opened 2020-05-01
Upgrading dependencies

schottra:
There's a security advisory for `acorn`, which is a dependency of `jest`. I've upgraded our versions of `jest`, `typescript`, and a few other libraries. This caused some errors due to types changing slightly or becoming a little more strict, and I address all of those as well.

* `codeCoverage` was turned off in `jest.config.js` because we explicitly enable it when running the yarn script to generate coverage and it causes errors when used with watch mode.
* Changed `test:coverage` to `test-coverage` so that it would not run as part of `yarn test`, but only when specifically invoked.


yarn.lock
```diff
@@ -3389,24 +3534,24 @@ acorn-globals@^4.1.0:
...
   version "6.2.0"
   resolved "https://registry.yarnpkg.com/acorn-walk/-/acorn-walk-6.2.0.tgz#123cb8f3b84c2171f1f7fb252615b1c78a6b1a8c"
   integrity sha512-7evsyfH1cLOCdAzZAd43Cic04yKydNx0cF+7tiA19p1XnLLPU4dpCQOqpjqwokFe//vS0QqfqqjCS2JkiIs0cA==
 
-acorn@^5.5.3:
-  version "5.7.4"
-  resolved "https://registry.yarnpkg.com/acorn/-/acorn-5.7.4.tgz#3e8d8a9947d0599a1796d10225d7432f4a4acf5e"
-  integrity sha512-1D++VG7BhrtvQpNbBzovKNc1FLGGEE/oGe7b9xJm/RFHMBeUaUGpluV9RLjZa47YFdPcDAenEYuq9pQPcMdLJg==
+acorn@^6.0.1, acorn@^6.0.7, acorn@^6.4.1:
+  version "6.4.1"
```
schottra, line 3547: This is security vulnerability being updated (acorn should be at 6.4.1 or later)

PR #69, opened 2020-05-05
feat: Adding automatic image building and release management

schottra:
* Pulled in the latest `boilerplate` to get some shared github actions.
* Began using the GH action to build/push images for pull requests
* Added a `semantic-release` config to allow automatic generation of release notes and releases when merging changes to master. See [semantic-release docs](https://github.com/semantic-release/semantic-release) for more info.
* Added commit linting to ensure that commits follow [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/) standards. This helps us with the automatic release generation process.
* fixed a test that was flaky on CI due to the expectation that an event loop would take at least 1ms to complete.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -1,0 +1,18 @@
+name: Pull Request
+on:
+  pull_request:
+
+jobs:
+  build_image:
+    name: Build Image
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Push Docker Image to Github Registry
+        uses: whoan/docker-build-with-cache-action@v5
+        with:
+          username: "${{ github.actor }}"
+          password: "${{ secrets.GITHUB_TOKEN }}"
+          image_name: ${{ secrets.package_name }}
+          image_tag: ${{ github.sha }}
+          registry: docker.pkg.github.com
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -1,6 +1,6 @@
 name: Pull Request
 on:
-  pull_request:
+  pull_request
 
 jobs:
   build_image:
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -1,6 +1,5 @@
 name: Pull Request
-on:
-  pull_request
+on: pull_request
 
 jobs:
   build_image:
@@ -16,3 +15,12 @@ jobs:
           image_name: ${{ secrets.package_name }}
           image_tag: ${{ github.sha }}
           registry: docker.pkg.github.com
+      - name: Push Docker Image to Dockerhub
+        uses: whoan/docker-build-with-cache-action@v5
+        with:
+          username: "${{ secrets.DOCKERHUB_USERNAME }}"
+          password: "${{ secrets.DOCKERHUB_PASSWORD }}"
+          image_name: ${{ secrets.DOCKERHUB_IMAGE_NAME }}
+          image_tag: ${{ github.sha }}
+          push_git_tag: true
+          build_extra_args: "--compress=true"
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -2,8 +2,8 @@ name: Pull Request
 on: pull_request
 
 jobs:
-  build_image:
-    name: Build Image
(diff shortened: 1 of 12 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,67 @@
...
+      - name: Install dependencies
+        run: yarn install --frozen-lockfile
+      - name: Release
+        env:
+          GITHUB_TOKEN: ${{ secrets.RELEASEBOT_TOKEN }}
+          GIT_AUTHOR_NAME: "service-github-lyft-semantic-release"
+          GIT_AUTHOR_EMAIL: "service-github-lyft-semantic-release@users.noreply.github.com"
+          GIT_COMMITTER_NAME: "service-github-lyft-semantic-release"
+          GIT_COMMITTER_EMAIL: "service-github-lyft-semantic-release@users.noreply.github.com"
+        run: npx semantic-release
```
EngHabu, line 27: I was hoping we can use an already available github action that does this.... but I guess that's ok

.github/workflows/release.yml
```diff
@@ -0,0 +1,67 @@
...
+        uses: actions/checkout@v2
+      - name: Setup Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: 12
+      - name: Install dependencies
+        run: yarn install --frozen-lockfile
+      - name: Release
+        env:
+          GITHUB_TOKEN: ${{ secrets.RELEASEBOT_TOKEN }}
```
EngHabu, line 22: Can't you use the current user's token?

schottra: We cannot. Since we use branch protection, the user needs permission to commit on master without an approval in order to perform the release function. The default github token will not have this permission, but the release bot does.
This should be relatively safe, as the secrets are not provided to forks of the repo, and only trusted users are allowed to create branches in the primary repo.

.github/workflows/release.yml
schottra: We could always publish this as a separate GH action to make it more reusable?

.github/workflows/release.yml
EngHabu: I was thinking from a fork standpoint, whether people who fork can set these variables to reasonable values that will make it work for them...

So you are saying even if the user has permission to perform, the default token github makes available doesn't have this "scope" in it? or that contributors of this repo won't? if it's the latter, then that's ok because the person who clicked "Squash and Merge" has the needed permissions to merge into master, so their token will have the same permissions. If it's the former, then sure, I understand.

.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,18 @@
...
+on:
+  pull_request
+
+jobs:
+  build_image:
+    name: Build Image
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Push Docker Image to Github Registry
```
EngHabu, line 11: ah, I also started pushing to Dockerhub (again) because it turns out you can't pull github docker images unless you authenticate, making the onboarding process a bit more complicated... so for now I'm pushing to both, and using github docker registry for testing (when an image is needed) and dockerhub for lyft/flyte references (as is the case today)

.github/workflows/release.yml
schottra: Ah yeah so this isn't set up to allow forks to publish releases.
I'm not certain that the action runs as the user who merges. I think it runs as the `github-actions` account? I _think_ that account does have write access but is not considered an Admin (and thus can't push directly to master). I'd prefer to merge this as-is for now and then explore ways to make it compatible with forks. It may require having all of the git credentials specified as repository secrets to get it working correctly.

.github/workflows/pull_request.yml
schottra: Do you have an example of that somewhere? As in, is there an action I can already consume to do the push to dockerhub as well?

schottra: Nevermind, I found it in flytepropeller. I updated the places we are pushing to also push to dockerhub.

.github/workflows/release.yml
EngHabu: yeah I think that would be cool.. this also updates the code right? not just the github release? if you ever get to make it a github action I would also make it output the new version so that we can simplify the code below...

PR #70, opened 2020-05-20
feat: adds more identifying information for node executions

schottra:
# Execution Details Page
* Show the task / workflow name below each node in the NodeExecutions list
* Show the task / workflow name below the header in the DetailsPanel
* Updates formatting logic for items in the Executions tab of the DetailsPanel to only show attempt numbers. This is in line with some upcoming changes to the hierarchy in the NodeExecutionsList
* Unit tests for the above changes

image

image


.editorconfig
```diff
@@ -20,7 +20,7 @@ insert_final_newline = true
 [*.json]
 indent_size = 2
 
-[*.yaml]
+[*.{yaml,yml}]
 indent_size = 2
```
BobNisco: Is there any benefit to maintaining both file-extension specific configs here and also in `prettier`? It looks like prettier handles more file types compared to this file, and given prettier's ability to format files on-save in an IDE, or in the pre-commit hook that's set up here, I'm not sure I understand the utility of keeping file-extension specific configs in this file.

tsconfig.json
```diff
@@ -1,28 +1,28 @@
 {
```
schottra, line 1: This was a result of running `yarn lint-fix` manually on one of my commits, which processed all JSON files in the repo to get them in line with our `prettier` config :-)

.editorconfig
schottra: (The base for this PR is branched from another PR which I haven't merged yet. I did a rebase/force push/base change to avoid showing that diff here. But I'll answer the question anyway)
`.editorconfig` is just a convenience for people who use editors which don't support tslint. it will at least give them a better chance at getting close to our syntax without having to wait for the commit lint to yell at them.
So no it's not necessarily required. But since it's already set up, I decided to update it 😄

PR #72, opened 2020-06-02
refactor: fetch child node executions for NodeExecution rows

schottra:
One of the end goals of our dynamic task experience changes is that instead of the hierarchy in the table being `NodeExecution -> TaskExecution/WorkflowExecution -> NodeExecution`, we remove the intermediate entity and make it `NodeExecution -> NodeExecution`.

This creates some issues:
* To determine if a NodeExecution can be expanded, we need to know if it has any children. There is no property on a NodeExecution that tells us that it has children. To determine that information, we need different behavior based on the node type.
  * For Task nodes (dynamic tasks), we will need to fetch the TaskExecutions created by the given NodeExecution. For each TaskExecution, we need to fetch its child NodeExecutions.
  * For Workflow nodes (specifically those that launch a separate WorkflowExecution), we need to fetch all NodeExecutions which are children of the given WorkflowExecution identifier.
* For some types of Nodes, the children need to be grouped by retry attempt. At the moment, this only applies to Task nodes, but it's useful for the concept of "groups of children" to be generic across all types of Nodes, so that the rendering logic is reusable. Workflow nodes don't currently allow retries, but may in the future. So they will only ever yield one group of children.

To address those issues, I updated the rendering/fetching logic as follows:
* Added a `useChildNodeExecutions` hook which will implement the necessary logic to fetch and group children for a given NodeExecution, varying the logic based on the type of node.
  * For Task nodes, we fetch all of the task executions, and then the node executions for each of those
  * For workflow nodes, we fetch the NodeExecution children for the workflow execution id.
  * The RequestConfig is piped through from the top level and used for all nested requests for NodeExecutions. This means that any filtering/sorting will apply to all levels of the list.
* Updated `NodeExecutionRow` to only render an expander if the value returned from `useChildNodeExecutions` is an array with length greater than zero.
* Refactored and exported some of the work functions for fetching `NodeExecutions` and `TaskExecutions` so that they can be composed by other hooks.

### DIFF
--- a/src/components/Executions/ExecutionDetails/contexts.ts
+++ b/src/components/Executions/ExecutionDetails/contexts.ts
@@ -1,6 +1,7 @@
 import * as React from 'react';
 
-import { Execution, NodeExecution } from 'models';
+import { Execution, NodeExecution, RequestConfig } from 'models';
+import { ExecutionFiltersState } from '../filters/useExecutionFiltersState';
 
 export interface ExecutionContextData {
     execution: Execution;
@@ -11,4 +12,8 @@ export const ExecutionContext = React.createContext<ExecutionContextData>(
 );
 export const NodeExecutionsContext = React.createContext<
     Dictionary<NodeExecution>
+>({});
+
+export const NodeExecutionsRequestConfigContext = React.createContext<
+    RequestConfig
 >({});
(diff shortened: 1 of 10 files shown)


src/components/Executions/ExecutionDetails/contexts.ts
```diff
@@ -1,6 +1,7 @@
 import * as React from 'react';
 
-import { Execution, NodeExecution } from 'models';
+import { Execution, NodeExecution, RequestConfig } from 'models';
+import { ExecutionFiltersState } from '../filters/useExecutionFiltersState';
```
podehnal, line 4: I don't see `ExecutionFiltersState` being used anywhere in this file, is that intentional?

PR #75, opened 2020-06-18
feat: Remove intermediate NodeExecutionsTable row content

schottra:
_TODOs and additional cleanup will happen in a separate PR_

* Implemented `ExecutionDataCache`, which is a shared data store used by components in the `NodeExecutionsTable`. This facilitates looking up the various entities (Nodes, Tasks, Workflows, WorkflowExecutions) that are needed at each level of the NodeExecutionsTable. This cache is provided via a context which sits above all components in the tree.
* Updated all hooks/components in the `ExecutionDetails` and `TaskExecutionDetails` pages to load their data using an `ExecutionDataCache`, so that all related entities are stored and available in nested components.
* Update the code for mapping `NodeExecution` -> `DetailedNodeExecution` (`useDetailedNodeExecution[s]`) to be pure functional transforms instead of working with `Fetchable`s. The fetching/waiting logic has been moved up to the consuming components/hooks.
* Moved contexts used in Execution components up to a common folder
* Moved concept of `DetailedNodeExecution` down into the `NodeExecutionsTable`, since components above that do not need the extra information.
* Updated `NodeExecutionChildren` to render "child groups" of `NodeExecution`s, instead of rendering either `TaskExecution` children or `WorkflowExecution` children. We still have differentiated behavior for the two cases, but both of them are now returning groups of `NodeExecution`s. 
  * The primary use case for this is to group child `NodeExecutions` by attempt of the parent.
  * A previous PR implemented the basics of fetching child `NodeExecution`s. This one is making use of that functionality to actually render the children.
* Moved column definitions for the `NodeExecutionsTable` into shared context. We don't need to be generating these definitions at every level of the table.
* Fixed some flaky tests that have started consistently failing

Before:
image

After:
image


src/components/Executions/Tables/NodeExecutionChildren.tsx
```diff
@@ -1,132 +1,50 @@
...
-    RefreshConfig,
-    useDataRefresher,
-    useWorkflowExecution
-} from 'components/hooks';
-import { isEqual } from 'lodash';
-import { Execution, NodeExecutionClosure, WorkflowNodeMetadata } from 'models';
+import { Typography } from '@material-ui/core';
+import * as classnames from 'classnames';
+import { useExpandableMonospaceTextStyles } from 'components/common/ExpandableMonospaceText';
 import * as React from 'react';
```
podehnal, line 4: nit: prefer to import "external" (ie react) packages before "local" ones and before components.

src/components/Executions/Tables/NodeExecutionChildren.tsx
```diff
@@ -1,132 +1,50 @@
...
-    );
-    useTaskExecutionsRefresher(nodeExecution, taskExecutions);
-
-    const columnStyles = useColumnStyles();
-    // Memoizing columns so they won't be re-generated unless the styles change
-    const { columns, Skeleton } = React.useMemo(() => {
-        const columns = generateTaskExecutionColumns(columnStyles);
-        return { columns, Skeleton: generateRowSkeleton(columns) };
-    }, [columnStyles]);
+    const showNames = childGroups.length > 1;
```
podehnal, line 17: nit: prefer this logic inline if only being used once on line 30

src/components/Executions/Tables/NodeExecutionRow.tsx
```diff
@@ -1,49 +1,58 @@
...
+        requestConfig,
+        workflowExecution
+    });
+
+    const detailedChildNodeExecutions = useDetailedChildNodeExecutions(
+        childNodeExecutions.value
     );
 
-    const isExpandable = childNodeExecutions.value.length > 0;
+    const isExpandable = detailedChildNodeExecutions.length > 0;
```
podehnal, line 48: nit: prefer this logic inline if only used on line 57

src/components/Executions/Tables/contexts.ts
```diff
@@ -1,6 +1,14 @@
...
+export interface NodeExecutionsTableContextData {
+    columns: NodeExecutionColumnDefinition[];
+    state: NodeExecutionsTableState;
+}
 
 export const NodeExecutionsTableContext = React.createContext<
-    NodeExecutionsTableState
->({} as NodeExecutionsTableState);
+    NodeExecutionsTableContextData
+>({} as NodeExecutionsTableContextData);
```
BobNisco, line 14: While I see that this follows the code that existed here before this change, should we consider giving this a default implementation instead?

src/components/Executions/contexts.ts
```diff
@@ -16,3 +18,7 @@ export const NodeExecutionsContext = React.createContext<
 export const NodeExecutionsRequestConfigContext = React.createContext<
     RequestConfig
 >({});
+
+export const ExecutionDataCacheContext = React.createContext<
+    ExecutionDataCache
+>({} as ExecutionDataCache);
```
BobNisco, line 24: Same comment here as above

src/components/Executions/useChildNodeExecutions.ts
```diff
@@ -1,91 +1,100 @@
...
         )
     );
+
+    // Remove any empty groups
+    return groups.reduce<NodeExecutionGroup[]>((out, group) => {
+        if (group === null || group.nodeExecutions.length === 0) {
+            return out;
+        }
+        return [...out, group];
+    }, []);
```
BobNisco, line 92: Feels like a filter is a more applicable function to use here

```suggestion
    return groups.filter((group) => {
        return group !== null && group.nodeExecutions.length;
    }, []);
```

src/components/Executions/utils.ts
```diff
@@ -114,96 +115,81 @@ export const taskExecutionIsTerminal = (taskExecution: TaskExecution) =>
...
+    const nodeInfo = dataCache.getNodeForNodeExecution(nodeExecution.id);
 
-    if (workflow) {
-        if (!workflow.closure) {
-            throw new Error('Workflow has no closure');
+    let displayId = nodeId;
+    let displayType = NodeExecutionDisplayType.Unknown;
+    let taskTemplate: TaskTemplate | undefined = undefined;
+
+    if (nodeInfo == null) {
```
BobNisco, line 131: Is the double-equal syntax here expected? Or can we utilize `===`

src/components/hooks/useTask.ts
```diff
@@ -11,22 +10,20 @@ import { FetchableData } from './types';
...
-            // for us. If we get to this point, something went wrong.
-            doFetch: () =>
-                Promise.reject(
-                    new NotFoundError(
-                        'Task template',
-                        'No template has been loaded for this task'
-                    )
-                )
+            doFetch: async taskId =>
+                (await getTask(taskId)).closure.compiledTask.template
```
BobNisco, line 26: Do we have to handle any potential errors from the `await getTask` call?

src/components/Executions/Tables/NodeExecutionChildren.tsx
schottra: I would normally agree. I'm using a lint rule/vscode command to enforce the ordering on imports. I think this was the closest I can get, and I had a preference for automated fixing and consistency across all contributors/files. :-/

src/components/Executions/Tables/NodeExecutionChildren.tsx
schottra: That makes sense. I use this pattern so that it's clear _what_ the logic is being used for: determining if we show the names of the groups. When placed inline, it feels like more cognitive load to figure out that the names are conditionally shown _and_ to figure out what controls it.
But it's probably a matter of preference?

src/components/Executions/Tables/NodeExecutionRow.tsx
schottra: Same answer as above: trying to make it clear that there's a logical dependency.

src/components/Executions/Tables/contexts.ts
schottra: I debated this with myself. Providing a default value gives a fallback for scenarios where a value isn't explicitly provided. These contexts are intended to _only_ be provided in the view where they are being used, effectively scoping the "caching" to only that instance of the view and creating a new context every time.
I'm concerned that failing to provide a scoped context will result in unintended side effects (high memory usage, stale entries) because the objects live around across multiple instantiations of the view.
So, the intention here is that if it is used incorrectly, it will be quite obvious by the errors thrown while attempting to access properties of an empty object.

Alternatively, I could provide a default value that throws/warns upon access so that it's clear that you've hit the default context when that was not expected?

src/components/Executions/useChildNodeExecutions.ts
schottra: Eek. I agree. I'm not sure why this is a reduce and not a filter. Let me refactor it.

src/components/Executions/utils.ts
schottra: `==` is checking for null _or_ undefined. I agree that it is not exactly obvious.

src/components/hooks/useTask.ts
schottra: Nope, this is all within a promise chain and `useFetchable` will handle any errors and bubble them up to a failed state.

src/components/Executions/Tables/contexts.ts
schottra: Thinking about it some more, and reading some reasoning from others on the same idea, I'm going to leave it as-is and allow the resulting errors to serve as a warning that there needs to be a useful value provided somewhere in the tree for these contexts.

src/components/Executions/contexts.ts
schottra: Resolving with the same answer as above.

PR #78, opened 2020-06-30
fix: Expose cluster information for Workflow Executions

schottra:
The primary goal was to expose a piece of metadata that is useful in debugging. But since the portion of the page where we show execution details is already pretty crowded, I decided to implement a restyling that would be part of some upcoming design changes for the execution details page. This gave me more room to add the extra piece of information

* Moved most details out of header into a separate section below the header, which is collapsible
* Updated execution name to render all as one string, making it easier to copy
* Updated styling of details
* Added 'Cluster' detail. Executions with no `spec.metadata.systemMetadata.executionClusterName` property will just show "(unknown)"
* Added errors for WorkflowExecutions, shown at the bottom of the metadata section

Before:

image

After:

image

Expansions:
ExpandCollapseExecutionMetadata

### DIFF
--- a/src/components/Executions/ExecutionDetails/ExecutionMetadata.tsx
+++ b/src/components/Executions/ExecutionDetails/ExecutionMetadata.tsx
@@ -5,16 +5,21 @@ import { unknownValueString } from 'common/constants';
 import { formatDateUTC, protobufDurationToHMS } from 'common/formatters';
 import { timestampToDate } from 'common/utils';
 import { useCommonStyles } from 'components/common/styles';
-import { secondaryBackgroundColor, smallFontSize } from 'components/Theme';
+import { secondaryBackgroundColor } from 'components/Theme';
 import { Execution } from 'models';
 import * as React from 'react';
+import { ExpandableExecutionError } from '../Tables/ExpandableExecutionError';
 import { ExecutionMetadataLabels } from './constants';
 
 const useStyles = makeStyles((theme: Theme) => {
     return {
         container: {
-            alignItems: 'center',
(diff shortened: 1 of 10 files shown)


src/components/Executions/ExecutionDetails/ExecutionMetadata.tsx
```diff
@@ -0,0 +1,100 @@
...
+    const commonStyles = useCommonStyles();
+    const styles = useStyles();
+
+    const { domain } = execution.id;
+    const { duration, startedAt, workflowId } = execution.closure;
+    const { systemMetadata } = execution.spec.metadata;
+    const cluster =
+        systemMetadata && systemMetadata.executionCluster
+            ? systemMetadata.executionCluster
+            : unknownValueString;
```
BobNisco, line 53: Could leverage optional chaining + nullish coalescing operator here

```suggestion
    const cluster = systemMetadata?.executionCluster ?? unknownValueString;
```

src/components/Executions/ExecutionDetails/constants.ts
```diff
@@ -0,0 +1,7 @@
+export const executionMetadataLabels = {
```
BobNisco, line 1: Could be an enum instead for better type support throughout where you use this

src/components/Executions/ExecutionDetails/ExecutionMetadata.tsx
schottra: Oh I hadn't realized null-coalescing was a thing in TS. Thanks!

PR #80, opened 2020-07-06
fix: prevents collapsing of execution errors when scrolled out of view

schottra:
lyft/flyte#64

Since we use virtualization, the components used to render the rows in this table are unmounted every time they leave the viewport. This is great for performance, but it resets state when the component comes back into view. To account for this, we're now storing the expansion state for errors at the table level, and piping them through as an initial state when mounting the components. This will maintain the expansion state when the component leaves and reenters the viewport.

Note: This change _should_ have a corresponding unit test. However, since jsdom does not support layout, we can't test scrolling the component out of the viewport and back in. So it's not possible at the moment to write a unit test for it :-(

Before: 
CollapsingError

After:
NonCollapsingError

### DIFF
--- a/src/components/Executions/Tables/WorkflowExecutionsTable.tsx
+++ b/src/components/Executions/Tables/WorkflowExecutionsTable.tsx
@@ -263,7 +263,7 @@ export const WorkflowExecutionsTable: React.FC<WorkflowExecutionsTableProps> = p
         const execution = executions[rowProps.index];
         const cacheKey = getCacheKey(execution.id);
         const onExpandCollapseError = (expanded: boolean) => {
-            setExpandedErrors({ ...expandedErrors, [cacheKey]: expanded });
+            setExpandedErrors(currentExpandedErrors => ({ ...currentExpandedErrors, [cacheKey]: expanded }));
             recomputeRow(rowProps.index);
         };
         return (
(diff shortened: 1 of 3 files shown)


src/components/Executions/Tables/WorkflowExecutionsTable.tsx
```diff
@@ -197,44 +260,21 @@ export const WorkflowExecutionsTable: React.FC<WorkflowExecutionsTableProps> = p
...
     // are in a failed state
     const rowRenderer: ListRowRenderer = rowProps => {
-        const { index: rowIndex, style } = rowProps;
-        const execution = executions[rowIndex];
-        const { error } = execution.closure;
-
+        const execution = executions[rowProps.index];
+        const cacheKey = getCacheKey(execution.id);
+        const onExpandCollapseError = (expanded: boolean) => {
+            setExpandedErrors({ ...expandedErrors, [cacheKey]: expanded });
```
BobNisco, line 266: We probably want to use the callback version of `setExpandedErrors` when merging objects to ensure we aren't utilizing stale state 

```suggestion
            setExpandedErrors(currentExpandedErrors => ({ ...currentExpandedErrors, [cacheKey]: expanded }));
```

PR #84, opened 2020-07-15
fix: show info message for un-launchable workflows

schottra:
lyft/flyte#291

This change improves the experience when a user is attempting to launch a Workflow for which it is impossible to specify a required value.
This can happen if the Workflow defines a required input with a type that the Console doesn't support yet (such as Schema, Binary, etc). If the Launch Plan specifies a default value, or we are re-launching an execution from which we can copy the value, it is still possible to proceed. Otherwise, we won't be able to successfully launch.
The previous experience was a little confusing, because we would only show a generic error at the bottom of the form after attempting to submit. The new experience is to show the user an error message indicating which inputs are unsupported, and directing them to either select a Launch Plan which provides a value, or to use the CLI to launch the workflow

image

### DIFF
--- a/src/components/Launch/LaunchWorkflowForm/inputHelpers/utils.ts
+++ b/src/components/Launch/LaunchWorkflowForm/inputHelpers/utils.ts
@@ -1,6 +1,6 @@
 import { Core } from 'flyteidl';
 import { get } from 'lodash';
-import { InputType } from '../types';
+import { InputType, InputTypeDefinition } from '../types';
 
 /** Performs a deep get of `path` on the given `Core.ILiteral`. Will throw
  * if the given property doesn't exist.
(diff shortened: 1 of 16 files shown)


src/components/Launch/LaunchWorkflowForm/useLaunchWorkflowFormState.ts
```diff
@@ -50,6 +51,8 @@ function useLaunchPlanSelectorOptions(launchPlans: LaunchPlan[]) {
     ]);
 }
 
+function useUnsupportedRequiredInputs() {}
```
BobNisco, line 54: This this used anywhere?

src/components/Launch/LaunchWorkflowForm/inputHelpers/utils.ts
```diff
@@ -25,3 +25,36 @@ export function collectionChildToString(type: InputType, value: any) {
...
+            if (!subtype) {
+                console.error(
+                    'Unexpected missing subtype for collection input',
+                    typeDefinition
+                );
+                return false;
+            }
+            return typeIsSupported(subtype);
+        }
+        default:
```
BobNisco, line 57: Depending on requirements here, we could also consider asserting never in the default block to make sure this switch covers all `InputType`s at compile time

src/components/Launch/LaunchWorkflowForm/useLaunchWorkflowFormState.ts
schottra: Oops 🤦

src/components/Launch/LaunchWorkflowForm/inputHelpers/utils.ts
schottra: Oh I like that. Thanks!

PR #86, opened 2020-07-20
feat: Enables support for Blobs on the Launch form

schottra:
lyft/flyte#378

This adds basic input support for blobs by rendering a micro form containing `uri`, `format` and `dimensionality` sub-inputs. The only required input from a user is the `uri` field. Dimensionality is technically required, but is implemented with a select control that defaults to a value. So it's not possible for a user to specific an empty dimensionality in the simple case.

Collections of Blobs are also supported, following the same rules as other input types (it is specified with a JSON array string). In this case, each value must include a `uri` and `dimensionality`.

`format` is always optional and must be a string.

image

image

### DIFF
--- a/src/components/Launch/LaunchWorkflowForm/BlobInput.tsx
+++ b/src/components/Launch/LaunchWorkflowForm/BlobInput.tsx
@@ -1,28 +1,133 @@
+import {
+    FormControl,
+    FormHelperText,
+    InputLabel,
+    MenuItem,
+    Select,
+    TextField,
+    Typography
+} from '@material-ui/core';
+import { makeStyles, Theme } from '@material-ui/core/styles';
 import { BlobDimensionality } from 'models';
 import * as React from 'react';
-import { InputProps } from './types';
+import {
+    blobFormatHelperText,
+    blobUriHelperText,
+    defaultBlobValue
+} from './constants';
+import { BlobValue, InputProps } from './types';
 import { getLaunchInputId } from './utils';
 
-/** A micro form for entering the values related to a Blob Literal
- */
+const useStyles = makeStyles((theme: Theme) => ({
+    dimensionalityInput: {
+        flex: '1 1 auto',
+        marginLeft: theme.spacing(2)
+    },
+    formatInput: {
+        flex: '1 1 auto'
+    },
+    inputContainer: {
+        borderLeft: `1px solid ${theme.palette.divider}`,
+        marginTop: theme.spacing(1),
+        paddingLeft: theme.spacing(1)
+    },
+    metadataContainer: {
+        display: 'flex',
+        marginTop: theme.spacing(1),
+        width: '100%'
+    }
+}));
+
+/** A micro form for entering the values related to a Blob Literal */
 export const BlobInput: React.FC<InputProps> = props => {
+    const styles = useStyles();
     const { error, label, name, onChange, value: propValue } = props;
-    const blobValue = typeof propValue === 'object' ? propValue : {};
+    const blobValue =
+        typeof propValue === 'object'
+            ? (propValue as BlobValue)
+            : defaultBlobValue;
     const hasError = !!error;
     // TODO: We might want a way to pass errors that are specific to a sub-field
     // of an input. Right now, a string error assumes that an input has a single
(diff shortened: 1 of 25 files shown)


src/components/Launch/LaunchWorkflowForm/BlobInput.tsx
```diff
@@ -0,0 +1,129 @@
...
+    }
+}));
+
+/** A micro form for entering the values related to a Blob Literal */
+export const BlobInput: React.FC<InputProps> = props => {
+    const styles = useStyles();
+    const { error, label, name, onChange, value: propValue } = props;
+    const blobValue = isObject(propValue)
+        ? (propValue as BlobValue)
+        : defaultBlobValue;
```
BobNisco, line 48: Suggestion to create a typeguard for `BlobValue` so we can avoid type casting. Something like:

```tsx
export function isBlobValue(data: unknown): data is BlobValue {
    return isObject(data);
}
```

And these lines would become:

```suggestion
    const blobValue = isBlobValue(propValue) ? propValue : defaultBlobValue;
```

src/components/Launch/LaunchWorkflowForm/BlobInput.tsx
```diff
@@ -0,0 +1,129 @@
...
+            format
+        });
+    };
+
+    const handleDimensionalityChange = ({
+        target: { value }
+    }: React.ChangeEvent<{ value: unknown }>) => {
+        onChange({
+            ...blobValue,
+            dimensionality: value as BlobDimensionality
```
BobNisco, line 75: Similar suggestion for type guards here if possible

src/components/Launch/LaunchWorkflowForm/inputHelpers/blob.ts
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// Allows for string values ('single'/'multipart') when specifying blobs manually in collections
+function getDimensionality(value: string | number) {
+    return typeof value === 'number'
+        ? value
+        : BlobDimensionality[
+              value.toUpperCase() as keyof typeof BlobDimensionality
+          ];
+}
```
BobNisco, line 33: Just a thought, we may be able to utilize more type guards here:

```tsx
export function isKeyOfBlobDimensionality(val: string): val is keyof typeof BlobDimensionality {
    return Object.keys(BlobDimensionality).indexOf(val) >= 0;
}

// Allows for string values ('single'/'multipart') when specifying blobs manually in collections
function getDimensionality(value: string | number) {
    if (typeof value === 'number') {
        return value;
    }
    if (isKeyOfBlobDimensionality(value)) {
        return BlobDimensionality[value];
    }
    // Note unclear what we would want to do in this situation
    return SOME_DEFAULT;
}
```

src/components/Launch/LaunchWorkflowForm/BlobInput.tsx
schottra: I'm not really sure what action we would take here upon checking the value and seeing that it's _not_ a dimensionality. It's coming from a `Select` component, not a text field. So it's not possible for it to be anything other than the values we provide to that component (which are type-checked elsewhere). This cast is because the type of the change event doesn't allow for specifying the type of the value.

src/components/Launch/LaunchWorkflowForm/inputHelpers/blob.ts
schottra: Typeguard is a good idea. I'm not sure what to do with a default either. The code which calls this function throws an exception if an invalid value is returned, so I don't think this function is expected to coerce an invalid value into a sensible default. Doing so might cause issues which are hard to diagnose by treating a URI specifying a directory as a single file and only processing the first file in the directory
So the preference is to completely fail if a bad value is provided.

schottra: It turns out there are two places which call this code. One does assume the value has already been validated. But either way, the result of indexing into an enum with a string that is _not_ a key is that it will return the first item in the enum. (Thanks, Typescript?) So I can preserve that behavior.

PR #88, opened 2020-08-12
fix: only filter NodeExecutions when viewing the Nodes tab

schottra:
lyft/flyte#425

We share the loaded NodeExecutions between the two tabs on the ExecutionDetails page.
This is a useful performance optimization when both are viewing an unfiltered list of NodeExecutions. But it results in missing node information when a filter is applied in the Nodes tab and then the Graph tab is opened.

To address this, I updated the logic in `ExecutionNodeViews` to only pass the filter list if the current tab is the Nodes tab. This has the effect of resetting the fetchable for NodeExecutions when the tab is changed if a filter is applied. But if no filter is applied, the passed list will be empty in all cases, resulting in reuse of the previously-fetched data.

The majority of this PR is actually cleanup work done to facilitate writing a unit test to cover the new behavior 😁

### DIFF
--- a/src/components/hooks/useTabState.ts
+++ b/src/components/hooks/useTabState.ts
@@ -1,7 +1,7 @@
 import { useState } from 'react';
 
 export function useTabState(
-    tabs: { [k: string]: string },
+    tabs: { [k: string]: string | object },
     defaultValue: string
 ) {
     const [value, setValue] = useState(defaultValue);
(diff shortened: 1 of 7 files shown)


src/components/hooks/useTabState.ts
```diff
@@ -1,7 +1,7 @@
 import { useState } from 'react';
 
 export function useTabState(
-    tabs: { [k: string]: string },
+    tabs: { [k: string]: string | object },
```
BobNisco, line 4: Should we type this?

```suggestion
    tabs: { [k: string]: string | { id: string; label: string } },
```

schottra: Good suggestion. At the moment, the shape of the object doesn't matter. So I'm okay leaving it generic.

PR #92, opened 2020-08-31
Get full inputs/outputs from execution data

kanterov:
# TL;DR
Fixes Input/Output view when using GCS.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description

flyteadmin doesn't know how to return public URLs for GCS objects. It returns URL in format `gs://...` that doesn't work.  This fixes Input/Output view when using GCS for Storage by using new fullInputs/fullOutputs fields in getExecutionData proto. Change is backward-compatible, if full inputs/outputs aren't present, it's going to fallback on the existing mechanism.

I had to take https://github.com/lyft/flyteconsole/commit/6a3b80f175b357bc81fd5f7fea31c8e89d13cd0d and few other commits because I needed new fields in flyteidl.

## Tracking Issue
https://github.com/lyft/flyte/issues/444

## Follow-up issue
_NA_

### DIFF
--- a/src/components/Executions/ExecutionDetails/ExecutionNodeDetails/OutputNodeDetails.tsx
+++ b/src/components/Executions/ExecutionDetails/ExecutionNodeDetails/OutputNodeDetails.tsx
@@ -15,7 +15,10 @@ const RemoteExecutionOutputs: React.FC<{ execution: Execution }> = ({
     return (
         <WaitForData {...executionData}>
             {() => (
-                <RemoteLiteralMapViewer blob={executionData.value.outputs} />
+                <RemoteLiteralMapViewer
+                    map={executionData.value.fullOutputs}
+                    blob={executionData.value.outputs}
+                />
             )}
         </WaitForData>
     );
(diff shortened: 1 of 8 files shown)


kanterov:
cc @kumare3

kumare3:
@kanterov today you showed that you know typescript 😜. But why not a separate PR

schottra:
@kanterov Thanks for doing this!

The other PR should be merged today and you can rebase off of master after that. Or you could just update the version of flyteidl in package.json and run `yarn` to update the lock file instead of pulling in all of the other changes.

Would you mind writing/updating tests for the components that you changed to make sure they use the new data when it's available and fall back to the old data otherwise?

kanterov:
@schottra done. Some non-code related builds checks are failing, it seems I need to configure something for them to work. Do I need to fix those?

schottra:
@kanterov I think the build/push actions are expected to fail, since they run from a fork and you won't have permissions to push to github/dockerhub. I should update those to not run on forked repositories.
However, the CLA check needs to be addressed before we merge. Would you mind clicking through on that and signing the contributor agreement if you haven't already?
If you have, then the email address associated with your commits also needs to be the one associated with your github account, so that the check can match them.

schottra:
@kanterov If you rebase on latest master, I've updated the Github actions, so two of those failed checks should now pass.

kanterov:
@schottra sorry it took some time, but now everything is done :) please take a look again

schottra (approved):

PR #96, opened 2020-09-09
refactor: make workflow details generic so it can be used for tasks

schottra:
lyft/flyte#459

Tasks/Workflows have a lot of similar information. The details view only really differs in whether or not we want to show schedules and what filter we need to use to fetch the executions.
So as a preparatory step for showing task details, this change moves all of the WorkflowDetails logic into a shared set of `Entity` components. Rendering the details for an entity only requires passing a `ResourceIdentifier`. The individual view components will be responsible for pulling the correct values from the route and rendering a `EntityDetails` component using the right `resourceType` value. 

*Note* While the work has been done to enable showing Task details, it isn't hooked up yet. That will be in a follow-up PR.

* Added `components/Entities` folder to hold components which are common across entity types (workflow, launch plan, task, etc..)
* Moved the components in `WorkflowDetails` into the new components folder
* Made `EntityDetails` components generic based on a passed `ResourceIdentifier`
  * Sections in the view are conditionally shown based on a set of flag constants which are mapped for each resourceType. The default view is just to show the description of the entity. Workflows will show everything. Tasks will show everything except schedules.
* Created maps for string constants and executions filters based on resource type.


src/components/Entities/EntityDetails.tsx
```diff
@@ -0,0 +1,97 @@
...
+export const EntityDetails: React.FC<EntityDetailsProps> = ({ id }) => {
+    const sections = entitySections[id.resourceType];
+    const project = useProject(id.project);
+    const styles = useStyles();
+    const [showLaunchForm, setShowLaunchForm] = React.useState(false);
+    const onLaunch = () => setShowLaunchForm(true);
+    const onCancelLaunch = () => setShowLaunchForm(false);
+
+    return (
+        <>
```
BobNisco, line 55: Nit: is the fragment necessary?

BobNisco (approved):

src/components/Entities/EntityDetails.tsx
schottra: Oh hm, good catch. It's probably not necessary. My only thought is that it might have to do with some obnoxious flexbox positioning stuff in the container above this. I didn't look closely at it when porting this component over.
I'll remove it in the next PR if it turns out not to be needed. If it _is_ needed, I'll be sure to document it or see if I can come up with a better solution.

src/components/Entities/EntityDetailsHeader.tsx
```diff
@@ -3,11 +3,12 @@ import { makeStyles, Theme } from '@material-ui/core/styles';
 import ArrowBack from '@material-ui/icons/ArrowBack';
 import * as classnames from 'classnames';
 import { useCommonStyles } from 'components/common/styles';
-import { Project } from 'models';
+import { Project, ResourceIdentifier } from 'models';
 import { getProjectDomain } from 'models/Project/utils';
 import * as React from 'react';
 import { Link } from 'react-router-dom';
 import { Routes } from 'routes';
+import { launchStrings } from './constants';
```
podehnal, line 11: Nit: I have always put constants like this in `SCREAMING_CAMEL` case but not sure what others do

podehnal: Nevermind this is a helper method :)

podehnal (commented):
👍

src/components/Entities/EntityDetailsHeader.tsx
schottra: Fair point to make!
I don't have a strong preference, especially in TS where more information is available.
But for this repo, the pattern is mostly to not use a screaming case. So I try to maintain that consistency.

PR #100, opened 2020-10-05
refactor: Make launch state and components generic

schottra:
This is a refactoring step to prepare the launch state machine and components for being able to handle tasks as a source.
I separated the state machine into a base config that is composed into more specific machines for workflow and task launching.

The form components and state hook were also made generic and now accept the xstate state/service generated by interpreting either a workflow or task machine. The base `LaunchForm` component will now conditionally render the correct form component based on the props passed to it.


podehnal (approved):
👍

PR #102, opened 2020-10-06
Cleanup work for launching single task executions.

schottra:
* Updated `EntityDetails` to pass the correct props when launching a Task
* Updated `RelaunchExecutionForm` and underlying hooks/functions to generate the correct initial parameters based on whether the source execution was a single task or a workflow.
* Updated header of Execution Details page to render the correct parent link based on what type of execution we are viewing (workflow vs. single task)
* Updated Relaunch flow to set the `referenceExecution` value when relaunching
* Updated ExecutionMetadata component to show referenceExecution values when they exist.


podehnal (approved):
👍

PR #107, opened 2020-10-20
feat: render CronSchedule

honnix:
# TL;DR
Render CronSchedule with optional `offset` value.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue
https://github.com/lyft/flyte/issues/529

## Follow-up issue
_NA_

### DIFF
--- a/src/common/formatters.ts
+++ b/src/common/formatters.ts
@@ -122,6 +122,31 @@ export function fixedRateToString({ value, unit }: Admin.IFixedRate): string {
     return `Every ${value} ${fixedRateUnitStrings[unit]}`;
 }
 
+const hourlyAliases = ['@hourly', 'hourly', 'hours'];
+const dailyAliases = ['@daily', 'daily', 'days'];
+const weeklyAliases = ['@weekly', 'weekly', 'weeks'];
+const monthlyAliases = ['@monthly', 'monthly', 'months'];
+const yearlyAliases = ['@yearly', 'yearly', 'years', '@annually', 'annually'];
+
+export function getScheduleFrequencyStringFromAlias(schedule: string) {
+    if (hourlyAliases.includes(schedule)) {
+        return 'Every hour';
+    }
+    if (dailyAliases.includes(schedule)) {
+        return 'Every day';
+    }
+    if (weeklyAliases.includes(schedule)) {
+        return 'Every week';
+    }
+    if (monthlyAliases.includes(schedule)) {
+        return 'Every month';
+    }
+    if (yearlyAliases.includes(schedule)) {
+        return 'Every year';
+    }
+    return '';
+}
+
 export function getScheduleFrequencyString(schedule?: Admin.ISchedule) {
     if (schedule == null) {
         return '';
@@ -135,7 +160,11 @@ export function getScheduleFrequencyString(schedule?: Admin.ISchedule) {
         return fixedRateToString(schedule.rate);
     }
     if (schedule.cronSchedule && schedule.cronSchedule.schedule) {
-        return cronstrue.toString(`${schedule.cronSchedule.schedule}`);
+        return (
+            getScheduleFrequencyStringFromAlias(
+                schedule.cronSchedule.schedule
+            ) || cronstrue.toString(schedule.cronSchedule.schedule)
+        );
     }
     return '';
 }
--- a/src/common/formatters.ts
+++ b/src/common/formatters.ts
@@ -93,6 +93,54 @@ export function millisecondsToHMS(valueMS: number): string {
     return parts.length ? parts.join(' ') : unknownValueString;
 }
 
+/** Outputs a value in moment.Duration in (Y M D H M S) format (ex. 1y 1M 1d 2h 3m 30s) */
+export function durationToYMWDHMS(duration: moment.Duration): string {
+    if (duration.asSeconds() === 0) {
+        return '';
+    }
+
+    const parts = [];
+
+    if (duration.years() !== 0) {
+        parts.push(`${Math.abs(duration.years())}y`);
+    }
+
+    if (duration.months() !== 0) {
+        parts.push(`${Math.abs(duration.months())}M`);
+    }
+
+    // ISO-8601 does not permit mixing between the PnYnMnD and PnW formats.
+    // Any week-based input is multiplied by 7 and treated as a number of days.
+    // However moment can parse the mixture resulting both a number of weeks and a number of days.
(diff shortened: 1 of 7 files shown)


honnix:
@schottra PTAL.

src/components/Launch/SchedulesTable.tsx
```diff
@@ -54,6 +57,18 @@ export const schedulesTableColumns: KeyedColumnProps[] = [
         label: 'frequency',
         width: schedulesTableColumnsWidths.frequency
     },
+    {
```
honnix, line 60: Not exactly sure how this is used.

schottra: This component isn't actually used at the moment. It's left over from the first iteration of the UI which had a very basic view of the schedules (both active and inactive) in a workflow.

src/common/formatters.ts
```diff
@@ -134,6 +134,19 @@ export function getScheduleFrequencyString(schedule?: Admin.ISchedule) {
     if (schedule.rate) {
         return fixedRateToString(schedule.rate);
     }
+    if (schedule.cronSchedule && schedule.cronSchedule.schedule) {
+        return cronstrue.toString(`${schedule.cronSchedule.schedule}`);
```
schottra, line 138: Is this string guaranteed to be in the correct format? Above, we're adding a leading 0 to ensure `cronstrue` can parse it correctly.

src/common/test/formatters.spec.ts
```diff
@@ -160,6 +163,41 @@ describe('millisecondsToHMS', () => {
     );
 });
 
+describe('getScheduleFrequencyString', () => {
```
schottra, line 166: Thank you for adding tests!

src/components/Entities/EntitySchedules.tsx
```diff
@@ -23,7 +26,11 @@ const RenderSchedules: React.FC<{
             {launchPlans.map((launchPlan, idx) => {
                 const { schedule } = launchPlan.spec.entityMetadata;
                 const frequencyString = getScheduleFrequencyString(schedule);
-                return <li key={idx}>{frequencyString}</li>;
+                const offsetString = getScheduleOffsetString(schedule);
+                const scheduleString = offsetString
+                    ? `${frequencyString} (offset: ${offsetString})`
```
schottra, line 31: I have some concerns about the way we are displaying this.

1. Based on the tests above, it seems like this might render as something like "Every day (offset: P1D)". Personally, that feels confusing to read because I have a plainly readable string and then have to mentally parse an offset string and add it in my head to figure out when it will actually run. Is there a way to make this work the way a plain cronstring does and render the whole thing, with offset, as a readable string?
2. The UI currently renders time-related values primarily as UTC, potentially with a local time in a muted text (see the executions table). This seems like it will render a value in an arbitrary time zone?

src/common/formatters.ts
schottra: Also, we don't need the formatting (`${}`) if we're just passing the value along?

honnix: At least in flytekit we verify it is a valid Unix flavoured cron expression. I noticed the padding 0 above to deal with AWS cron but that is not assured either and mostly replies on contract. Maybe a better way is to handle the exception and display something indicating invalid cron? Now I realize it could be an alias too like `daily`.

src/components/Entities/EntitySchedules.tsx
honnix: I will think about how to make it human readable. The offset is a duration so it's relative and not time zone bound. Together with cron, offset could be used in the scenario for example: execute it daily but against yesterday's data partition if offset is P-1D.

src/common/formatters.ts
schottra: Valid point about it just being convention/contract. I think my question was whether or not the expected value _should_ be a normal cron string based on convention. If we're ensuring that in flytekit, that's equivalent to the existing case.

src/components/Entities/EntitySchedules.tsx
schottra: I think I don't fully understand the way this cron expression is used. Offset seems to carry a different semantic meaning than simple arithmetic on the time value. I had interpreted offset to be a UTC offset.
But it sounds like it has extra logic attached to it if it is controlling what data will be used for input. In that case, I don't know if it makes sense to apply the offset to the schedule and render the resulting value.

Either way, I think this might get confusing for users. But if the way you are displaying it here makes sense to you as a user, I'm okay with shipping it as is and thinking about ways to make it more understandable.

src/common/formatters.ts
honnix: I will add additional logic to handle aliases.

src/components/Entities/EntitySchedules.tsx
honnix: Yeah I totally agree this might seem odd and require more understanding of what offset is. I'm OK not to render offset for now until we figure out a better way. Internally we are integrating flyteadmin directly with a [UI system](https://github.com/spotify/backstage), so this is not critical.

src/common/formatters.ts
```diff
@@ -93,6 +93,54 @@ export function millisecondsToHMS(valueMS: number): string {
     return parts.length ? parts.join(' ') : unknownValueString;
 }
 
+/** Outputs a value in moment.Duration in (Y M D H M S) format (ex. 1y 1M 1d 2h 3m 30s) */
+export function durationToYMWDHMS(duration: moment.Duration): string {
```
honnix, line 97: I had to re-implement this instead of refactoring the above function. Although yearly offset is not common, monthly and daily are.

honnix:
@schottra I fixed the `offset` rendering according to what we agreed. PTAL again, thanks.

schottra (approved):

schottra:
@honnix Looks good. Thanks for implementing this!

src/components/Launch/constants.ts
```diff
@@ -7,5 +7,6 @@ export const launchPlansTableColumnWidths = {
 export const schedulesTableColumnsWidths = {
     active: 80,
     frequency: 300,
+    offset: 300,
```
honnix, line 10: I made a mistake here. :( This is not needed.

PR #111, opened 2020-10-22
fix: don't skip legacy children check when isParentNode=false

schottra:
lyft/flyte#438

In implementing the original changes to use the `isParentNode` flag, I had assumed that the flag would be completely absent. But due to the way protobuf is deserialized, when the metadata object is present, `isParentNode` will default to `false`. This will cause us to incorrectly skip checking for children on legacy node executions by querying the child TaskExecutions first.

So, the logic is now to query directly for child node executions if the flag is true, and fall back to the legacy check otherwise.


EngHabu (approved):
Thank you!

PR #115, opened 2020-11-05
fix: stale selected item reference in NE details panel

schottra:
# TL;DR
Fixes an issue where status of a NodeExecution doesn't update in the DetailsPanel if you have the panel open when the status changes.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?
 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
We were assigning the selected item for the Details Panel using an object reference, instead of an id reference. When the node executions update, new copies of the items are returned. But the DetailsPanel was still rendering the previously referenced object. To fix it, I updated the state tracking the selected item to be by id / cache key and am now computing the object to return using a memoized `find` call on the array of executions whenever the array or the selected id change.
Also added a test to cover the regression.

## Tracking Issue
https://github.com/lyft/flyte/issues/505

## Follow-up issue
_NA_

### DIFF
--- a/src/components/Executions/Tables/useNodeExecutionsTableState.ts
+++ b/src/components/Executions/Tables/useNodeExecutionsTableState.ts
@@ -28,9 +28,7 @@ export function useNodeExecutionsTableState({
 
     const setSelectedExecution = useMemo(
         () => (newValue: DetailedNodeExecution | null) =>
-            setSelectedExecutionKey(
-                newValue == null ? null : newValue.cacheKey ?? null
-            ),
+            setSelectedExecutionKey(newValue?.cacheKey ?? null),
         [setSelectedExecutionKey]
     );
 
--- a/src/components/Executions/Tables/useNodeExecutionsTableState.ts
+++ b/src/components/Executions/Tables/useNodeExecutionsTableState.ts
@@ -14,10 +14,25 @@ export function useNodeExecutionsTableState({
         [value]
     );
 
-    const [
-        selectedExecution,
-        setSelectedExecution
-    ] = useState<DetailedNodeExecution | null>(null);
+    const [selectedExecutionKey, setSelectedExecutionKey] = useState<
+        string | null
+    >(null);
+
+    const selectedExecution = useMemo(
+        () =>
+            executions.find(
+                ({ cacheKey }) => cacheKey === selectedExecutionKey
+            ) || null,
+        [executions, selectedExecutionKey]
+    );
+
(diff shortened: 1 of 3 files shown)


src/components/Executions/Tables/useNodeExecutionsTableState.ts
```diff
@@ -14,10 +14,25 @@ export function useNodeExecutionsTableState({
...
-    ] = useState<DetailedNodeExecution | null>(null);
+    const [selectedExecutionKey, setSelectedExecutionKey] = useState<
+        string | null
+    >(null);
+
+    const selectedExecution = useMemo(
+        () =>
+            executions.find(
+                ({ cacheKey }) => cacheKey === selectedExecutionKey
+            ) || null,
```
BobNisco, line 25: If `.find` doesn't find a match it will return `undefined`. Curious to know if there is a meaningful difference between returning `undefined` vs `null`?

src/components/Executions/Tables/useNodeExecutionsTableState.ts
```diff
@@ -14,10 +14,25 @@ export function useNodeExecutionsTableState({
...
+        [executions, selectedExecutionKey]
+    );
+
+    const setSelectedExecution = useMemo(
+        () => (newValue: DetailedNodeExecution | null) =>
+            setSelectedExecutionKey(
+                newValue == null ? null : newValue.cacheKey ?? null
```
BobNisco, line 32: Nit: I believe this can be written as:

```suggestion
                newValue?.cacheKey ?? null
```

src/components/Executions/Tables/useNodeExecutionsTableState.ts
schottra: Short answer: No, not really. The consuming code should handle the undefined case in the same manner as the null case.

Longer answer: It's a matter of preference.

I prefer using `null` when explicitly setting a value to the "not defined" case and letting it be `undefined` when it truly was never defined or set. There are also cases where updating a value to be `undefined` is a little difficult or error-prone (like merging objects or attempting to `delete` properties). Setting to `null` feels cleaner in these cases.

To reflect that, the type of the returned value in the state is `DetailedNodeExecution | null`. It's initialized to `null` and set to `null` when the selection is cleared. So allowing it to be undefined here would cause a type error.

BobNisco (approved):

PR #117, opened 2020-11-06
fix: handle null LiteralMap in RemoteLiteralMapViewer

kanterov:
# TL;DR
Handle null LiteralMap in RemoteLiteralMapViewer

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
Not sure I completely understand how proto objects translate to JS objects. There are cases when `fullInputs` and `fullOutputs` are equal to null. Pull request adds a null check that complements the existing undefined check.

### DIFF
--- a/src/components/Literals/RemoteLiteralMapViewer.tsx
+++ b/src/components/Literals/RemoteLiteralMapViewer.tsx
@@ -28,7 +28,7 @@ const BlobTooLarge: React.FC<{ url: string }> = ({ url }) => (
  */
 export const RemoteLiteralMapViewer: React.FC<{
     blob: UrlBlob;
-    map?: LiteralMap;
+    map: LiteralMap | null;
 }> = ({ blob, map }) => {
     if (!blob.url || !blob.bytes) {
         return (
@@ -38,7 +38,7 @@ export const RemoteLiteralMapViewer: React.FC<{
         );
     }
 
-    if (map !== undefined) {
+    if (map != null) {
         return <LiteralMapViewer map={map} />;
     }
 
--- a/src/components/Executions/ExecutionDetails/test/RelaunchExecutionForm.test.tsx
+++ b/src/components/Executions/ExecutionDetails/test/RelaunchExecutionForm.test.tsx
@@ -84,7 +84,9 @@ describe('RelaunchExecutionForm', () => {
         task = createMockTask('MyTask');
         executionData = {
             inputs: { url: 'http://somePath', bytes: long(1000) },
-            outputs: {}
+            outputs: {},
+            fullInputs: null,
+            fullOutputs: null
         };
 
         mockGetWorkflow = jest.fn().mockResolvedValue(workflow);
--- a/src/components/Executions/Tables/__stories__/NodeExecutionsTable.stories.tsx
+++ b/src/components/Executions/Tables/__stories__/NodeExecutionsTable.stories.tsx
@@ -47,7 +47,13 @@ const nodeRetryAttempts = {
 
 const apiContext = mockAPIContextValue({
     getExecution: () => Promise.resolve(workflowExecution),
-    getNodeExecutionData: () => Promise.resolve({ inputs: {}, outputs: {} }),
+    getNodeExecutionData: () =>
+        Promise.resolve({
+            inputs: {},
+            outputs: {},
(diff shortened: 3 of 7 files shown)


src/components/Literals/RemoteLiteralMapViewer.tsx
```diff
@@ -38,7 +38,7 @@ export const RemoteLiteralMapViewer: React.FC<{
         );
     }
 
-    if (map !== undefined) {
+    if (map !== undefined && map !== null) {
```
schottra, line 41: If you're handling both cases, a coercing equality against `null` will do it and is a common pattern in this repository. 
`if (map != null) { ... }`

Generally when checking protobuf fields which are not primitive values, I use this form of check.

schottra:
@kanterov 
Sorry you got caught by this. You're correct that it's a quirk of the way protobuf objects are deserialized. You can look into the details in the [flyteidl generated JS code](https://github.com/lyft/flyteidl/blob/master/gen/pb-js/flyteidl.js). But that file is huge and github can't render it, so here's the relevant code for the response object that results in this null value:
```
admin.WorkflowExecutionGetDataResponse = (function() {

            /**
             * Properties of a WorkflowExecutionGetDataResponse.
             * @memberof flyteidl.admin
             * @interface IWorkflowExecutionGetDataResponse
             * @property {flyteidl.admin.IUrlBlob|null} [outputs] WorkflowExecutionGetDataResponse outputs
             * @property {flyteidl.admin.IUrlBlob|null} [inputs] WorkflowExecutionGetDataResponse inputs
             * @property {flyteidl.core.ILiteralMap|null} [fullInputs] WorkflowExecutionGetDataResponse fullInputs
             * @property {flyteidl.core.ILiteralMap|null} [fullOutputs] WorkflowExecutionGetDataResponse fullOutputs
             */

            /**
             * Constructs a new WorkflowExecutionGetDataResponse.
             * @memberof flyteidl.admin
             * @classdesc Represents a WorkflowExecutionGetDataResponse.
             * @implements IWorkflowExecutionGetDataResponse
             * @constructor
             * @param {flyteidl.admin.IWorkflowExecutionGetDataResponse=} [properties] Properties to set
             */
            function WorkflowExecutionGetDataResponse(properties) {
                if (properties)
                    for (let keys = Object.keys(properties), i = 0; i < keys.length; ++i)
                        if (properties[keys[i]] != null)
                            this[keys[i]] = properties[keys[i]];
            }

            /**
             * WorkflowExecutionGetDataResponse outputs.
             * @member {flyteidl.admin.IUrlBlob|null|undefined} outputs
             * @memberof flyteidl.admin.WorkflowExecutionGetDataResponse
             * @instance
             */
            WorkflowExecutionGetDataResponse.prototype.outputs = null;

            /**
             * WorkflowExecutionGetDataResponse inputs.
             * @member {flyteidl.admin.IUrlBlob|null|undefined} inputs
             * @memberof flyteidl.admin.WorkflowExecutionGetDataResponse
             * @instance
             */
            WorkflowExecutionGetDataResponse.prototype.inputs = null;

            /**
             * WorkflowExecutionGetDataResponse fullInputs.
             * @member {flyteidl.core.ILiteralMap|null|undefined} fullInputs
             * @memberof flyteidl.admin.WorkflowExecutionGetDataResponse
             * @instance
             */
            WorkflowExecutionGetDataResponse.prototype.fullInputs = null;
/* ... */
```

The generated code to debug protobuf messages uses prototype inheritance to handle values which are absent in the protobuf message. It follows specific conventions for each data type, some of which cause problems if they aren't handled correctly.
For instance, booleans default to `false` (not undefined) if they weren't set. 
Nested objects will default to `null` instead of `undefined`.
Integers and enums will default to `0`.

I've debated potentially writing a transform layer that makes the models a little more reliable and less error-prone.

schottra:
@kanterov Looks like there is a type issue with one of the storybook stories due to updating `ExecutionData` to have non-optional properties. I'm happy to merge the PR once that is fixed!

kanterov:
@schottra thanks for your help, I fixed it, didn't have enough time yesterday :)

schottra (approved):

PR #127, opened 2020-12-04
test: adds setup for mock-service-worker

schottra:
This is the first step in using [`mock-service-worker`](https://mswjs.io) to mock our data instead of the messy mocking at the `apiContext` or `axios` level.

The basic idea is that we have a set of mock objects defined in `src/mocks` and a set of handler functions for each of the endpoints we use (there are only a handful). There will be a default set of objects that are loaded into the mock server for simple uses cases. Specific cases can add additional single-test handlers via `mockServer.use`. I decorated the mock server object with `insert*` functions for adding handlers for a specific entity type, to reduce the clutter in tests.

This PR just gets `mock-service-worker` integrated with a few simple objects and fixes whatever broke in doing so (mostly around some tests of the `AdminEntity` code).

A note:
Going forward, we will _not_ test things like whether the correct query params were passed to an API call. We _will_ test what happens in the failure cases. If we really need to test a specific behavior on query params, MSW can be set up to handle the query params and return different data. And then we can allow the component to render the data in the special case and assert that it looks the way we expect

### Change list
* Added `mock-service-worker`
* Updated the way our env variables are read to remove the unnecessary `isServer()` check. We can prefer `process.env` if it exists and otherwise try `window.env`.
* Added `handlers` for `WorkflowExecution`, `NodeExecution`, `Workflow` and `Project`.
  * Note: Handlers all use a shared response function which will encode the response to protobuf and set the correct headers, since the client code in AdminEntity will expect that and attempt to decode it. This is about as realistic a response as we can get.
* Added one mock object for a WF Execution and a small list of sample projects.
* Added code to import and create handlers for the default data objects
* Updated and/or removed some tests which were attempting to mock `AdmintEntity`. We will let them call all the way through to axios and operate data returned from our mock server.
* Updated some import statements of `AdminEntity` to be more specific, since the index-type includes cause circular reference issues.
* Moved around some constants for api path generation to make sure that the mock server and client code calling it will use the same prefix.
* Added some temporary empty tests to make Jest happy until I refactor the suites in question in my next PR.
* Upgraded `typescript` to fix error with types from msw, `jest`, `ts-jest`, and `babel-jest` to versions which support TS 4+. Fixed minors issues resulting from the upgrade.


src/components/Executions/ExecutionDetails/test/ExecutionNodeViews.test.tsx
```diff
@@ -29,6 +29,8 @@ describe.skip('ExecutionNodeViews', () => {
         props = { execution: workflowExecution };
     });
 
+    it('is disabled', () => {});
```
BobNisco, line 32: Are these intentionally kept in?

BobNisco, line 32: Are these intentionally kept in?

schottra: Yep. This keeps the test runner from throwing errors. I will remove them in the next PR when I fix and re-enable the commented-out tests.

BobNisco (approved):

podehnal (approved):

PR #128, opened 2020-12-08
Migrate from TSLint to ESLint

schottra:
Some of the changes in this feature branch fail linting checks because our updated version of Typescript is not compatible with the version of TSLint we are using.
However, TSLint has been officially deprecated. So instead of updating it, it seems like a better idea to just do the migration to ESLint with the Typescript plugins.

This migration was performed by running `npx tslint-to-eslint-config --prettier`, trimming down the ESLint config to extend recommended defaults and use the plugins we need for Typescript, Jest, Prettier, and React/Hooks, and then fixing all errors or disabling the rules generating them.
Note: There are still a lot of warnings from ESLint. These can be addressed later.

### Full change list

* Removed `tslint` and associated plugins, packages, config files.
* Upgraded `eslint` and added all necessary plugins
* Added ESLint config and ignore files
* Updated `package.json` scripts to use `eslint` instead of `tslint`
* Ran `eslint --fix` to auto-fix any simple errors based on the new settings
* Manually fixed any remaining errors or disabled/downgraded the rules in cases where we don't want them to be errors.


package.json
```diff
@@ -12,9 +12,8 @@
     "build": "TS_NODE_COMPILER_OPTIONS='{\"module\": \"commonjs\"}' webpack --config webpack.config.ts --mode=development",
     "clean": "rm -rf dist",
     "codecov": "codecov -f .coverage/coverage-final.json",
-    "lint": "tslint --project .",
-    "lint-fix": "tslint --fix --project . && prettier --write './**/*.{js,json}'",
-    "lint:hooks": "eslint \"src/**/*.{ts,tsx}\"",
+    "lint": "eslint . --ext .js,.jsx,.ts,.tsx",
+    "lint-fix": "eslint . --ext .js,.jsx,.ts,.tsx --fix && prettier --write './**/*.{js,json}'",
```
BobNisco, line 16: Do we also want prettier to format `ts` and `tsx` files?

BobNisco (approved):

package.json
schottra: Good catch. The separate call to prettier is unnecessary. ESLint is configured to use prettier for formatting rules, so `eslint --fix` should catch all of it.
I cleaned up the scripts and `lint-staged` config to skip json/jsx(we don't have jsx files) as well.

PR #130, opened 2020-12-16
refactor: clean up mock fixtures and re-enable tests for executions

schottra:
This PR fills out the mock support for executions and implements the remaining unit tests that needed refactoring.

* Refactored mock server into the `createAdminServer` module. This uses a simple map-based store with keys being derived based on the entity type, url, and query string parameters. This mimics the way react-query keys its fetch operations (and in fact uses the same mechanism for generating query keys). This makes the server able to return proper 404 responses for test cases where we actually want an object to be missing, and gives more flexibility in the way we implement our mock response handlers. It also allows for using a shared endpoint handler for each entity type, instead of registering middleware matching an exact full url for every single object we want to be able to return.
* Added helper functions for generating fully-specified mock entities. These use a default mock object and allow deep-merging of any properties that need to be overridden. This dramatically reduces the amount of code needed to create new mock cases for our complex objects (Workflow, Task, Execution, etc).
* Added helper functions for generating related entities such as an `Execution` based on a specific `Workflow`, a `TaskExecution` for a specific task, etc. This makes it easier to generate consistent objects which correctly imitate the way data works in a production environment.
* Added `insertFixture` function which will insert records into our mock server for all of the various entities in a test fixture object.
* _Deleted_ the generic mock data objects that I was cobbling together into test cases in favor of using test fixtures.
* Pared down the default data we insert to only be the projects list. Each individual test is now responsible for inserting its own data.
* Added an implementation of `stableStringify` for use in hashing queries in the mock server.
* Added helper function to find an ancestor by role type. I was repeating this logic in several places and it's not trivial.
* Refactored some of the path generation helper functions so that they can also be used in the mock server code.
* Added a `fixtures` folder containing logic to generate all the entities necessary for various use cases we need to test. These objects can be inserted into the mockServer with the `insertFixture` helper function.
* Updated some components to not use the `WaitForQuery` helper component and just return `null` if there is no data present. This saves an extra component in the tree and avoids a remount when the data is refreshing.
* Fixed broken refresh query logic in `useExecutionNodeViewsState`
* Re-implemented tests for `ExecutionNodeViews`, `NodeExecutionDetails`, `NodeExecutionsTable`
* Removed global query listeners for extracting child data objects and inserting them into the cache. This was not the right approach, as it was causing child items to be updated _any_ time the parent query changed state. Instead, we now pass the queryClient to the functions which generate the parent queries and allow them to only extract and store the child objects when the parent is successfully fetched.
* Added a type alias for the input to `getCacheKey`
* Added `DeepPartial` helper to add better type support for merge operations on mock objects.
* Updated react-query out of beta versions
* Disabled an eslint rule that was making mock exporting difficult


src/components/Executions/ExecutionDetails/useExecutionNodeViewsState.ts
```diff
@@ -26,11 +28,18 @@ export function useExecutionNodeViewsState(
     };
 
     const shouldEnableQuery = (executions: NodeExecution[]) =>
-        every(executions, nodeExecutionIsTerminal) &&
-        executionIsTerminal(execution);
+        !executionIsTerminal(execution) ||
+        some(executions, ne => !nodeExecutionIsTerminal(ne));
```
podehnal, line 32: nit: any reason to use the JS native `ARRAY.some`?

src/test/setupTests.ts
```diff
@@ -1,9 +1,16 @@
...
 import { mockServer } from 'mocks/server';
+import { obj } from './utils';
 
 beforeAll(() => {
+    insertDefaultData(mockServer);
     mockServer.listen({
-        onUnhandledRequest: 'error'
+        onUnhandledRequest: (req) => {
+            const message = `Unexpected request: ${obj(req)}`;
+            console.error(message);
```
podehnal, line 11: is this `console.error` statement expected?

podehnal (approved):
Looks good to me, just a couple of non blocking nits. Great stuff!

src/test/setupTests.ts
schottra: Oh I guess not. The thrown error will end up in the test output. I'll remove it.

src/components/Executions/ExecutionDetails/useExecutionNodeViewsState.ts
schottra: Only that I wasn't aware it existed :-)

PR #139, opened 2021-01-13
chore: Clean up TODOs and unused code.

schottra:
* Audited the existing `TODO`s in the code base, removed all obsolete/won't-fix, and added tracking issues for the remaining.
* Took care of any lint errors in files that were touched.
* Removed dead code/routes around old views that are no longer used


podehnal (approved):
🧹 👍

PR #140, opened 2021-01-14
chore: removing unused exports

schottra:
I ran ts-prune on the repo and cleaned up most of the items which were identified as unused exports. A lot of them are leftover components/utilities from refactored code.
There will be an additional PR coming to remove `index` files from folders in favor of direct imports from files, and that will result in some additional unused exports being removed.


BobNisco (approved):

podehnal (approved):
🧹 💨

PR #148, opened 2021-02-23
fix: Use name if preferredUsername is empty

EngHabu:
# TL;DR
Use `name` from `/me` response if `preferred_username` is null.

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
Not all OIdP providers fill in all user info fields. Google, we found, uses the `name` field but not the `preferred_username` field.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/657

### DIFF
--- a/src/components/Navigation/UserInformation.tsx
+++ b/src/components/Navigation/UserInformation.tsx
@@ -26,7 +26,11 @@ export const UserInformation: React.FC<{}> = () => {
                 {profile.value == null ? (
                     <LoginLink />
                 ) : (
-                    profile.value.preferredUsername
+                    profile.value.preferredUsername == null ? (
+                            profile.value.name
+                        ) : (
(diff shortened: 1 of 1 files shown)


src/components/Navigation/UserInformation.tsx
```diff
@@ -26,7 +26,11 @@ export const UserInformation: React.FC<{}> = () => {
                 {profile.value == null ? (
                     <LoginLink />
                 ) : (
-                    profile.value.preferredUsername
+                    profile.value.preferredUsername == null ? (
```
kumare3, line 29: one thing with typescript is need to check `undefined` at times.

kumare3 (approved):

PR #157, opened 2021-04-05
feat: add filter by username for executions

Professional0321:
This PR is for https://github.com/flyteorg/flyte/issues/671.

- Added new button that is responsible for filtering executions by user.
- Hide it in case that user not logged in
- When user log in, his own executions will be displayed by default and can switch to `all executions` by clicking filtering button.

Onlymyexecutions


src/components/Executions/filters/useUserByFilterState.ts
```diff
@@ -0,0 +1,53 @@
...
+            ? profile.value.name
+            : profile.value?.preferredUsername;
+
+    const [value, setValue] = useState(username);
+    const button = useFilterButtonState();
+
+    useEffect(() => {
+        if (button.open) setValue('');
+        else setValue(username);
+    }, [button.open]);
```
schottra, line 32: I think we may want to use `active` and `setActive` to track if the user has enabled the filter. And I don't think we need a separate button state.

```ts
// This is tracking whether the user has enabled the filter.
// Depending on what default behavior you want, setting this to true will either 
// show all executions or only show executions belonging to the current user. 
// I think based on the logic below, true === show executions for the current user.
const [active, setActive] = useState(true); 
const getFilter = () => {
        return active && userName
            ? [
                  {
                      value: userName,
                      key: queryStateKey,
                      operation: FilterOperation.EQ
                  }
              ]
            : [];
    };
```

src/components/Executions/filters/useUserByFilterState.ts
```diff
@@ -0,0 +1,53 @@
+import { FilterOperationName } from 'models/AdminEntity/types';
+import { useEffect, useState, useContext } from 'react';
+import { FilterByUserState } from './types';
+import { useFilterButtonState } from './useFilterButtonState';
+import { APIContext } from 'components/data/apiContext';
+
+interface FilterByUserStateArgs {
+    queryStateKey: string;
```
schottra, line 8: `queryStateKey` is used by filters to keep the value of the filter in sync with the URL in the address bar. You don't appear to be using this anywhere. So I would say either take it out, or implement the query state functionality.
[Here is how it is done in the SingleFilter](https://github.com/flyteorg/flyteconsole/blob/master/src/components/Executions/filters/useSingleFilterState.ts#L41)
I think you would want a query state key something like `currentUserOnly` and its value would be a `boolean`.

src/components/Executions/filters/useUserByFilterState.ts
```diff
@@ -0,0 +1,53 @@
...
+    filterOperation?: FilterOperationName;
```
schottra, line 9: I don't think we need a filter operation to be passed. The filter we are generating is always going to use `FilterOperation.EQ`.

src/components/Executions/ExecutionFilters.tsx
```diff
@@ -57,21 +91,41 @@ export const ExecutionFilters: React.FC<{ filters: FilterState[] }> = ({
     filters
 }) => {
     const styles = useStyles();
+    const { profile } = React.useContext(APIContext);
```
schottra, line 94: I made a suggestion below to move the hiding logic into the filter state. If you do that, we don't need user profile here at all.

src/components/Executions/filters/useUserByFilterState.ts
```diff
@@ -0,0 +1,53 @@
...
+}
+
+/** Maintains the state for a button to be used for filtering by user.
+ */
+export function useUserByFilterState({
```
schottra, line 14: Suggest renaming to `useCurrentUserOnlyFilterState`

src/components/Executions/ExecutionFilters.tsx
```diff
@@ -57,21 +91,41 @@ export const ExecutionFilters: React.FC<{ filters: FilterState[] }> = ({
...
-                        {...filter.button}
-                        active={filter.active}
-                        key={filter.label}
-                        onReset={filter.onReset}
-                        buttonText={filter.label}
-                        className={styles.filterButton}
-                        renderContent={renderContent}
-                    />
-                );
+                if (filter.type == 'byUser') {
```
schottra, line 99: * I think a Checkbox might make more sense for this interaction.
* Preference for `===` instead of `==`. We want strict equality for string checks.
* If we add the `hidden` field to filter state, we can simplify this a little bit.
* The preference in this code base is to return `null` instead of an empty fragment(`<></>`) so that React knows there is no work to do.

Those changes might look something like:
```tsx
{filters.map(filter => {
    if ( !!filter.hidden ) {
        return null;
    }
    if (filter.type === 'boolean') {
        const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => filter.setActive(event.target.checked);
        // Using a standard MaterialUI checkbox component. You may want to do this differently
        return (
            <FormControlLabel
                key={filter.label}
                control={
                    <Checkbox
                        checked={filter.active}
                        onChange={handleChange}
                     />
                }
                label={filter.label}
        />)
    }
    const renderContent = () => <RenderFilter filter={filter} />;
    return (
        <FilterPopoverButton
            {...filter.button}
            active={filter.active}
            key={filter.label}
            onReset={filter.onReset}
            buttonText={filter.label}
            className={styles.filterButton}
            renderContent={renderContent}
        />
    );
}
```

src/components/Executions/ExecutionFilters.tsx
```diff
@@ -11,20 +18,47 @@ import {
...
+            flexDirection: 'row',
+            height: theme.spacing(7),
+            minHeight: theme.spacing(7),
+            paddingLeft: theme.spacing(1),
+            width: '100%'
+        },
+        filterButton: {
+            marginLeft: theme.spacing(1)
+        },
+        button: {
```
schottra, line 39: If we _were_ using a button, these styles should all already exist as part of the theme. I don't think we would want to repeat all of them here. But as I said below, I think we want to use a checkbox.

src/components/Tables/filters/FilterByUserButton.tsx
```diff
@@ -0,0 +1,75 @@
+import { Button } from '@material-ui/core';
```
schottra, line 1: I think this component probably isn't necessary if you go the checkbox route. But if not, I would make this component a more generic one which handles rendering a `BooleanFilterState`

src/components/Executions/filters/useUserByFilterState.ts
```diff
@@ -0,0 +1,53 @@
...
+    queryStateKey,
+    filterOperation = FilterOperationName.EQ
+}: FilterByUserStateArgs): FilterByUserState {
+    const { profile } = useContext(APIContext);
```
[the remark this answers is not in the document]
Professional0321: What is the main purpose to use context here is to reduce api call.
As you said, we can use useUserProfile directly in every component, but also it will make a new api call every time.
So if we define profile in APIContext, we can also use profile in every component without making api call.

katrogan:
@Professional0321 how did you test this? I've observed that in an authenticated context I never see any of my executions, despite having just launched some.  Does the filter work end to end when authentication is enabled?

src/components/Executions/filters/useUserByFilterState.ts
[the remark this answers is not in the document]
schottra: `useUserProfile` has [caching set to true](https://github.com/flyteorg/flyteconsole/blob/master/src/components/hooks/useUserProfile.ts#L12) so you _shouldn't_ see multiple requests unless there is a bug in the useFetchable hook or something is causing it to force reload.
Are you seeing it issue multiple requests in the devtools?

src/components/Executions/filters/useCurrentUserOnlyFilterState.ts
```diff
@@ -0,0 +1,43 @@
+import { FilterOperationName } from 'models/AdminEntity/types';
+import { UserProfile } from 'models/Common/types';
+import { useState } from 'react';
+import { BooleanFilterState } from './types';
+import { useFilterButtonState } from './useFilterButtonState';
+import { useUserProfile } from 'components/hooks/useUserProfile';
+
+/** Maintains the state for a button to be used for filtering by user.
```
schottra, line 8: Small nit: I think this doc string goes above line 14 instead? That way, code editors which render documentation will show this string whenever a user hovers over a usage of `useCurrentUserOnlyFilterState`

src/components/Executions/filters/useCurrentUserOnlyFilterState.ts
```diff
@@ -0,0 +1,43 @@
...
+ */
+function getUserId(profile: UserProfile): string {
+    return profile.sub ? profile.sub : '';
```
schottra, line 11: This is different from what you were using before (checking preferredName and a different field). I'm not sure what `sub` is for or whether it's always the right value to use. @EngHabu do you know?

src/components/Executions/filters/useCurrentUserOnlyFilterState.ts
```diff
@@ -0,0 +1,43 @@
...
+    };
+
+    return {
+        active,
+        button,
+        hidden: !userId,
+        label: 'Only my executions',
+        getFilter,
+        setActive,
+        onReset: () => {},
```
schottra, line 40: I think this filter _could_ have a valid `onReset` which reverts the active status back to its default value (`true`).

schottra (commented):
This is looking good. Thanks for working on it. I have a few last concerns:
* I think based on the way the logic works, if the API call to fetch the user profile takes a while to return, we will end up loading the executions without the filter and then re-rendering and loading the executions _with_ the filter once we have the value. If that's a UX concern (@EngHabu ?), we might have to look at delaying either the rendering of the table/filters or the requesting of the executions until we are sure that the userProfile has loaded.
* Do you think you could write some tests for this code?
* Could you add a screenshot or gif to the PR description of what this looks like in the UI?
* Does the Storybook still run and work properly? I think there are some stories which render the executions table and filters.

src/components/Executions/filters/useCurrentUserOnlyFilterState.ts
Professional0321: Yeah, I used wrong value for filtering executions before. In this API request, we need `userId`, not `name`. In the above code, `sub` is the same with `userId`. 
So it is right value for filter.

EngHabu:
@Professional0321 is this ready for another review?

Professional0321:
> @Professional0321 is this ready for another review?

PR #158, opened 2021-04-13
Add filter username

Professional0321:
Added filter by username.

- Added checkbox button to filter logged user's own executions
- Added `useCurrentUserOnlyFilterState` to provide filter condition
- Improved `ProjectExecutions` test with filter based on username

filter_by_username


kumare3:
Looks good to me 👍🏽 . @schottra can you +1?

kumare3 (approved):

src/components/Entities/EntityExecutions.tsx
```diff
@@ -50,6 +50,11 @@ export const EntityExecutions: React.FC<EntityExecutionsProps> = ({ id }) => {
         }
     );
 
+    /** Don't render component until finish fetching user profile */
+    if (filtersState.filters[4].status !== 'LOADED') {
```
schottra, line 54: You could just use `useUserProfile` directly here and check the status on that. We don't need to add status to a filter and pipe it through.
All react-query queries for a particular piece of data are synchronized and will receive the same values.
```tsx
const { isLoading: userProfileIsLoading } = useUserProfile();
if (userProfileIsLoading) {
    return null;
}
```

src/components/Executions/ExecutionFilters.tsx
```diff
@@ -53,13 +58,37 @@ const RenderFilter: React.FC<{ filter: FilterState }> = ({ filter }) => {
...
+                }
+                if (filter.type === 'boolean') {
+                    const handleChange = (
+                        event: React.ChangeEvent<HTMLInputElement>
+                    ) => filter.setActive(event.target.checked);
+
+                    return (
+                        <FormControlLabel
+                            key={filter.label}
+                            data-testid="checkbox"
```
schottra, line 80: This is very generic. Test ids should be unique. But if we had multiple boolean filters being rendered, they would all have the same id.  Also, test id is supposed to be a last resort for getting a component when using react-testing-library, since it doesn't test what the user is seeing. We should be able to get this component by its label instead. [See here for more info](https://testing-library.com/docs/queries/about#priority)

src/components/Executions/filters/types.ts
```diff
@@ -20,13 +20,15 @@ export interface FilterButtonState {
...
-export type FilterStateType = 'single' | 'multi' | 'search';
+export type FilterStateType = 'single' | 'multi' | 'search' | 'boolean';
 
 export interface FilterState {
     active: boolean;
     button: FilterButtonState;
+    hidden?: boolean;
     label: string;
     type: FilterStateType;
+    status?: string;
```
schottra, line 31: This could be removed if you implement my suggestion above regarding querying the user profile directly in `EntityExecutions`

src/components/Executions/filters/useCurrentUserOnlyFilterState.ts
```diff
@@ -0,0 +1,46 @@
...
+    return {
+        active,
+        button,
+        hidden: !userId,
+        label: 'Only my executions',
+        getFilter,
+        setActive,
+        onReset,
+        type: 'boolean',
+        status: profile.state.value.toString()
```
schottra, line 44: Similar comment here, remove and just get user profile directly in `EntityExecutions`

src/components/Project/ProjectExecutions.tsx
```diff
@@ -95,10 +95,15 @@ export const ProjectExecutions: React.FC<ProjectExecutionsProps> = ({
...
     );
 
-    return (
-        <div className={styles.container}>
-            <ExecutionFilters {...filtersState} />
-            <ErrorBoundary>{content}</ErrorBoundary>
-        </div>
-    );
+    /** Don't render component until finish fetching user profile */
+    if (filtersState.filters[4].status === 'LOADED') {
```
schottra, line 99: Same recommendation here for using `useUserProfile` directly.

src/components/Project/test/ProjectExecutions.test.tsx
```diff
@@ -20,59 +20,116 @@ import {
...
             </QueryClientProvider>,
             { wrapper: MemoryRouter }
         );
 
-    it('displays executions after successful load', async () => {
-        const { getByText } = renderView();
-        await waitFor(() => expect(getByText(executions[0].id.name)));
+    it('should not display checkbox if user does not login', async () => {
+        const { queryByTestId } = renderView();
+        await waitFor(() => {});
```
schottra, line 99: Prefer not to use empty `waitFor` functions. You can put your assertions into the body of the function and the test will wait until they pass.
Also, you can check for the checkbox by label instead of using a test-id.
```ts
await waitFor(() => {
    expect(mockGetUserProfile).toHaveBeenCalled();
    expect(queryByLabel('Only my executions')).toBeNull();
});
```

src/components/Project/test/ProjectExecutions.test.tsx
```diff
@@ -20,59 +20,116 @@ import {
...
+        expect(mockGetUserProfile).toHaveBeenCalled();
+        expect(queryByTestId(/checkbox/i)).toBeNull();
+    });
+
+    it('should display checkbox if user login', async () => {
+        mockGetUserProfile.mockResolvedValue(sampleUserProfile);
+        const { queryByTestId } = renderView();
+        await waitFor(() => {});
```
schottra, line 107: Suggest:
```ts
const { getByLabelText } = renderView();
await waitFor(() => { // Will wait until the function doesn't throw or return falsy
  const checkbox = getByLabelText('Only my executions'); // Will throw if not found
  expect(checkbox).toBeChecked(); // jest-dom has a built-in helper for checkboxes
});
```

src/components/Project/test/ProjectExecutions.test.tsx
```diff
@@ -20,59 +20,116 @@ import {
...
-    [sortQueryKeys.key]: executionSortFields.createdAt
-};
+jest.mock('components/Executions/Tables/WorkflowExecutionsTable');
 
 describe('ProjectExecutions', () => {
     let basicPythonFixture: ReturnType<typeof basicPythonWorkflow.generate>;
     let failedTaskFixture: ReturnType<typeof oneFailedTaskWorkflow.generate>;
-    let executions: Execution[];
+    let executions1: Execution[];
+    let executions2: Execution[];
```
schottra, line 34: Can you give these more descriptive names? It's hard to tell what is different between them or what use case is being tested with each of them.

src/components/Project/test/ProjectExecutions.test.tsx
```diff
@@ -20,59 +20,116 @@ import {
...
     let scope: DomainIdentifierScope;
     let queryClient: QueryClient;
+    let mockGetUserProfile: jest.Mock<ReturnType<typeof getUserProfile>>;
+
+    const sampleUserProfile: UserProfile = {
+        sub: 'sub'
+    } as UserProfile;
+
+    const defaultQueryParams1 = {
```
schottra, line 43: Same comment here about the names.

src/components/Project/test/ProjectExecutions.test.tsx
```diff
@@ -20,59 +20,116 @@ import {
...
+        mockGetUserProfile.mockResolvedValue(sampleUserProfile);
+        const { getByText, queryByText, queryByTestId } = renderView();
+        await waitFor(() => {});
+        expect(mockGetUserProfile).toHaveBeenCalled();
+        const checkbox = queryByTestId(/checkbox/i)?.querySelector(
+            'input'
+        ) as HTMLInputElement;
+        expect(checkbox).toBeTruthy();
+        expect(checkbox?.checked).toEqual(true);
+        await waitFor(() =>
```
schottra, line 124: Also recommending here: 
* Put the assertion on line 118 into the `waitFor` above it.
* Use `getByLabelText` to get the checkbox
* Use `toBeChecked` to check if the checkbox is checked

PR #171, opened 2021-06-01
fix: handle fullInputs for relaunch workflow

jsonporter:
Fixes issue where recent changes made to `flyteadmin` were not being supported in UI. The use-case is when a user goes to relaunch a workflow, we now return raw/full inputs when the payload is small enough; this adds support for `fullInputs`

see: https://github.com/flyteorg/flyteadmin/commit/067dc911bae09ee73cfcc6d46a658aa9a1c1d7d1#diff-9b311b663f43eccf79220bfcae3a565ad26d83532c945f39cfc1f7ba029c22e3

### DIFF
--- a/src/components/Executions/ExecutionDetails/test/RelaunchExecutionForm.test.tsx
+++ b/src/components/Executions/ExecutionDetails/test/RelaunchExecutionForm.test.tsx
@@ -185,6 +185,44 @@ describe('RelaunchExecutionForm', () => {
         });
     });
 
+    describe('Launch form with full inputs', () => {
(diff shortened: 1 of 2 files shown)


schottra (dismissed):
Seems alright to me.
I know we don't have a lot of test coverage around this, but we do cover the case of fetching remote inputs here: https://github.com/flyteorg/flyteconsole/blob/master/src/components/Executions/ExecutionDetails/test/RelaunchExecutionForm.test.tsx
Would be nice if you can add a test case for full inputs there as well!

schottra (approved):

PR #172, opened 2021-06-07
fix: this is rfc, do not deploy until resolved; details in comments

jsonporter:
## RFC: DO NOT DEPLOY UNTIL RESOLVED

I am working on the fix for https://github.com/flyteorg/flyte/issues/1085 and have identified where the workflow breaks. This change fixes the issue (short version: id's were being used as keys but their values were truncated and thus we had duplicates) however I am new to the code base and unclear on the implications of removing this ternary.  Where/why do we use `metadata` rather than `id`? 

Can someone with more context please take a look and help me understand if there are any edge cases I am missing here?

## Type
 - [ X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [ ] Unit tests added

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1085

### DIFF
--- a/src/components/Executions/utils.ts
+++ b/src/components/Executions/utils.ts
@@ -112,8 +112,9 @@ export const taskExecutionIsTerminal = (taskExecution: TaskExecution) =>
     taskExecution.closure &&
     terminalTaskExecutionStates.includes(taskExecution.closure.phase);
 
+/** Returns a NodeId from a given NodeExecution  */
 export function getNodeExecutionSpecId(nodeExecution: NodeExecution): string {
-    return nodeExecution.metadata?.specNodeId || nodeExecution.id.nodeId;
+    return nodeExecution.id.nodeId;
 }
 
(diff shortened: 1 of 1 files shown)


schottra:
@jsonporter I could be wrong, but I believe this is a case of backwards-compatibility. For complex nested workflows, we may end up executing the same node multiple times. Each node _execution_ will have a dynamically-generated id. But it will point to a node in the workflow spec by using `metadata.specNodeId`. However, we aren't guaranteed that the specNodeId will be set, so we fall back to `node.id`. I believe by removing the usage of metadata, we would break certain use cases of dynamic execution where a node is run multiple times.
cc @kumare3 and @anandswaminathan for more context.

kumare3:
@jsonporter whats the status on this one?

Pianist038801 (approved):

PR #178, opened 2021-07-26
Support enum input type in launch form

Pianist038801:

# TL;DR
Add enum type input in launch form

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1217

## Follow-up issue
_NA_

### DIFF
--- a/src/components/Launch/LaunchForm/SimpleInput.tsx
+++ b/src/components/Launch/LaunchForm/SimpleInput.tsx
@@ -2,15 +2,24 @@ import {
     FormControl,
     FormControlLabel,
     FormHelperText,
+    MenuItem,
+    Select,
     Switch,
     TextField
 } from '@material-ui/core';
+import { makeStyles } from '@material-ui/core/styles';
 import * as React from 'react';
 import { DatetimeInput } from './DatetimeInput';
 import { makeStringChangeHandler, makeSwitchChangeHandler } from './handlers';
 import { InputProps, InputType } from './types';
 import { UnsupportedInput } from './UnsupportedInput';
 import { getLaunchInputId } from './utils';
+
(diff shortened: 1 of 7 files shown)


src/components/Launch/LaunchForm/SimpleInput.tsx
```diff
@@ -2,28 +2,43 @@ import {
...
 } from '@material-ui/core';
+import { makeStyles } from '@material-ui/core/styles';
 import * as React from 'react';
 import { DatetimeInput } from './DatetimeInput';
 import { makeStringChangeHandler, makeSwitchChangeHandler } from './handlers';
 import { InputProps, InputType } from './types';
 import { UnsupportedInput } from './UnsupportedInput';
 import { getLaunchInputId } from './utils';
 
+const useStyles = makeStyles(theme => ({
```
jsonporter, line 18: NIT: not a blocker but do we need makeStyles here; couldn't we just send styles as a generic object?  (is there a specific reason for useStyle?)

eg, 
```
const formStyle = {
   minWidth: 100%; 
}
```

jsonporter (commented):
Left one comment (NIT) but once addressed lgtm.

src/components/Launch/LaunchForm/SimpleInput.tsx
Pianist038801: If we define a style that way, we should use style props of React component.
But I found that we currently use className prop for styling.

That's why I went with such solution, just tried to replicate how it's already implemented.

jsonporter (approved):
Lgtm

PR #185, opened 2021-08-18
fix: duration is not always calculated for task executions

csirius:
# TL;DR

Fixed the `unknown` is being present for both `start time` and `duration` in task details panel for nodes.
Added `last updated` to be shown instead if both of above fields are absent in API response.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added
 - [x] Code documentation added

## Complete description
Created a `fallback` option when `started_at` and `duration` fields are not present in the API response.

## Tracking Issue

fixes https://github.com/flyteorg/flyte/issues/885

## Follow-up issue
_NA_


jsonporter (approved):
lgtm 👍

PR #199, opened 2021-09-10
fix: show field types on json launch form

Pianist038801:
This PR makes JSON Launch Form show field types like other inputs.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?
 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Tracking Issue
fixes https://github.com/flyteorg/flyte/issues/1419


src/components/Launch/LaunchForm/StructInput.tsx
```diff
@@ -21,6 +21,25 @@ muiTheme.typography.h5 = {
     fontWeight: 400
 };
 
+const formatJson = data => {
+    const keys = Object.keys(data);
+
+    if (keys.includes('title')) {
+        const { title, type, format } = data;
+        data['title'] = `${title} (${format ? format : type})`;
```
csirius, line 29: what about to use `${format ?? type}` instead?

src/components/Launch/LaunchForm/StructInput.tsx
```diff
@@ -21,6 +21,25 @@ muiTheme.typography.h5 = {
...
+        if (!keys.includes('additionalProperties')) return data;
```
csirius, line 30: What if we have other fields but don't have `additionalProperties`?
Just to make sure what this means on my end.

csirius (changes requested):
Could you let me know your ideas?

src/components/Launch/LaunchForm/StructInput.tsx
Pianist038801: That’s a good catch, David. I just forgot to replace it with '??' as I wrote separate .js file to write this function. When I used in .js file, it was throwing an error saying SyntaxError: Unexpected token '?'

Pianist038801: Will push a fix ASAP.

src/components/Launch/LaunchForm/StructInput.tsx
Pianist038801: 'additionalProperties' is all I can see from the api response for now. If there is anything more, we can cope with it later.

src/components/Launch/LaunchForm/StructInput.tsx
csirius: I think the error is thrown coz the js file was not using ES6 or later versions.
For ts, it's okay. :)

src/components/Launch/LaunchForm/StructInput.tsx
csirius: Yeah, we can do that later.

csirius (approved):
LGTM!

PR #215, opened 2021-09-23
fix: polish workflow details page

csirius:
Signed-off-by: csirius <85753828+csirius@users.noreply.github.com>

# TL;DR
Remove space between versions table and the bar chart and increase minimal height of execution list view in workflow details page.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added
 - [x] Code documentation added

## Complete description
- Remove space between versions table and bar chart when there are not enough versions
- Add some minimal height for executions list view

## Tracking Issue

fixes https://github.com/flyteorg/flyte/issues/1517

## Follow-up issue
_NA_

### DIFF
--- a/src/components/Entities/EntityDetails.tsx
+++ b/src/components/Entities/EntityDetails.tsx
@@ -33,12 +33,11 @@ const useStyles = makeStyles((theme: Theme) => ({
         flex: '1 1 auto',
         flexDirection: 'column',
         margin: `0 -${theme.spacing(contentMarginGridUnits)}px`,
-        minHeight: theme.spacing(55)
+        flexBasis: theme.spacing(80)
     },
     versionsContainer: {
         display: 'flex',
-        flexDirection: 'column',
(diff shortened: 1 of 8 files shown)


jsonporter (commented):
One of these changes breaks the graph; the graphing library complains it needs a width and a height.  I think its because of the change on [line 29](src/components/common/ContentContainer.tsx).

I think we can fix it by either keeping height (and maxHeight) or alternatively we can add a container dive to [ReactFlowWrapper](https://github.com/flyteorg/flyteconsole/blob/6c2126308f3bd20ec7d43324598baf692d72b5c5/src/components/flytegraph/ReactFlow/ReactFlowWrapper.tsx#L133)

kumare3 (approved):

PR #251, opened 2021-12-07
feat: Remove disable auth var

jsonporter:
This PR removes `DISABLE_AUTH`.

The reasoning behind this decision is that the app itself does not use this value and is simply pass along with requests to Admin as `withCredentials`. However, this value has consequence on the server side as it informs `Access-control`; specifically if cross domain cookies can be set which can cause issues when connecting to Admin (and has no affect otherwise). 

see: https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/withCredentials

Edit:
- This PR also includes a minor fix for displaying nested graphs

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
Removes `DISABLE_AUTH` from flyteconsole.

## Tracking Issue
_NA_

## Follow-up issue
_NA_

### DIFF
--- a/src/components/flytegraph/ReactFlow/ReactFlowGraphComponent.tsx
+++ b/src/components/flytegraph/ReactFlow/ReactFlowGraphComponent.tsx
@@ -16,7 +16,7 @@ const ReactFlowGraphComponent = props => {
         root: data,
         nodeExecutionsById: nodeExecutionsById,
         onNodeSelectionChanged: onNodeSelectionChanged,
-        maxRenderDepth: 2
+        maxRenderDepth: 1
     } as ConvertDagProps);
 
     const backgroundStyle = getRFBackground().nested;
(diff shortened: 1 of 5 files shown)


src/components/flytegraph/ReactFlow/ReactFlowGraphComponent.tsx
```diff
@@ -16,7 +16,7 @@ const ReactFlowGraphComponent = props => {
         root: data,
         nodeExecutionsById: nodeExecutionsById,
         onNodeSelectionChanged: onNodeSelectionChanged,
-        maxRenderDepth: 2
+        maxRenderDepth: 1
```
csirius, line 19: Is this one related to this PR?

csirius: Just curious

csirius (approved):

PR #253, opened 2021-12-14
feat: Add support for securityContext in launch form.

jsonporter:
This PR adds support for `securityContext` in Admin requests.  Currently this feature is not fully rolled out but this change will support both `authRole` and `securityContext` during the migration to prevent any breaking changes. The change will appear on Launch/Relaunch/Recover actions. This change also removes the either/or toggle between K8 and IAM roles - now users can enter both/either/none. 

see: https://github.com/flyteorg/flyteidl/blob/f4809c1a52073ec80de41be109bccc6331cdbac3/protos/flyteidl/core/security.proto#L111

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
FlyteAdmin is planning to migrate to using `securityContext` in the future however to support older versions of Admin we will send and receive both values for the time being.

## Tracking Issue
_NA_

## Follow-up issue
_NA_

### DIFF
--- a/src/components/Launch/LaunchForm/LaunchRoleInput.tsx
+++ b/src/components/Launch/LaunchForm/LaunchRoleInput.tsx
@@ -1,163 +1,25 @@
-import {
-    FormControl,
-    FormControlLabel,
-    FormLabel,
-    Radio,
-    RadioGroup,
-    TextField,
-    Typography
-} from '@material-ui/core';
-import { log } from 'common/log';
+import { TextField, Typography } from '@material-ui/core';
 import { NewTargetLink } from 'components/common/NewTargetLink';
-import { useDebouncedValue } from 'components/hooks/useDebouncedValue';
-import { Admin } from 'flyteidl';
 import * as React from 'react';
-import { launchInputDebouncDelay, roleTypes } from './constants';
+import { AuthRoleStrings } from './constants';
 import { makeStringChangeHandler } from './handlers';
-import { useInputValueCacheContext } from './inputValueCache';
 import { useStyles } from './styles';
-import { InputValueMap, LaunchRoleInputRef, RoleType } from './types';
+import { LaunchRoleInputRef, LaunchRoles } from './types';
+import { AuthRoleTypes } from './types';
 
-const roleHeader = 'Role';
-const roleDocLinkUrl =
-    'https://github.com/flyteorg/flyteidl/blob/3789005a1372221eba28fa20d8386e44b32388f5/protos/flyteidl/admin/common.proto#L241';
-const roleTypeLabel = 'type';
-const roleInputId = 'launch-auth-role';
-const defaultRoleTypeValue = roleTypes.iamRole;
+const docLink =
+    'https://github.com/flyteorg/flyteidl/blob/f4809c1a52073ec80de41be109bccc6331cdbac3/protos/flyteidl/core/security.proto#L111';
 
 export interface LaunchRoleInputProps {
-    initialValue?: Admin.IAuthRole;
+    initialValue?: any;
     showErrors: boolean;
 }
-
-interface LaunchRoleInputState {
-    error?: string;
-    roleType: RoleType;
-    roleString?: string;
-    getValue(): Admin.IAuthRole;
-    onChangeRoleString(newValue: string): void;
-    onChangeRoleType(newValue: string): void;
-    validate(): boolean;
-}
-
-function getRoleTypeByValue(value: string): RoleType | undefined {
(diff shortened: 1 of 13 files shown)


csirius (dismissed):
LGTM!

src/components/Launch/LaunchForm/constants.ts
```diff
@@ -18,21 +12,21 @@ export const formStrings = {
...
+//         label: 'IAM Role',
+//         value: 'assumableIamRole'
+//     },
+//     k8sServiceAccount: {
+//         helperText: 'example: default-service-account',
+//         inputLabel: 'service account name',
+//         label: 'Kubernetes Service Account',
+//         value: 'kubernetesServiceAccount'
+//     }
+// };
```
csirius, line 29: Do we need this later?

src/components/Launch/LaunchForm/LaunchRoleInput.tsx
```diff
@@ -109,7 +109,9 @@ export const LaunchRoleInputImpl: React.RefForwardingComponent<
                     </Typography>
                 </div>
                 <TextField
-                    id={`launchform-role-${AuthRoleStrings[AuthRoleTypes.IAM]}`}
+                    id={`launchform-role-${
+                        AuthRoleStrings[AuthRoleTypes.IAM].inputLabel
+                    }`}
```
csirius, line 114: I think we should eliminate whitespaces from the `id` field.
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id#:~:text=id%20's%20value%20must%20not,have%20one%20single%20ID%20value

src/components/Launch/LaunchForm/LaunchRoleInput.tsx
```diff
@@ -126,7 +128,9 @@ export const LaunchRoleInputImpl: React.RefForwardingComponent<
                     </Typography>
                 </div>
                 <TextField
-                    id={`launchform-role-${AuthRoleStrings[AuthRoleTypes.k8]}`}
+                    id={`launchform-role-${
+                        AuthRoleStrings[AuthRoleTypes.k8].inputLabel
+                    }`}
```
csirius, line 133: The same applies to here.

src/components/Launch/LaunchForm/test/LaunchTaskForm.test.tsx
```diff
@@ -398,234 +425,73 @@ describe('LaunchForm: Task', () => {
...
+                it('for IAM', async () => {
+                    const { getByText, getByLabelText } = renderForm();
+                    await waitFor(() =>
+                        getByLabelText(
+                            AuthRoleStrings[AuthRoleTypes.IAM].inputLabel,
+                            {
+                                exact: false
+                            }
+                        )
+                    );
```
csirius, line 438: What's this for?
Are we missing `expect` or is this for some other purpose?

src/components/Launch/LaunchForm/test/LaunchTaskForm.test.tsx
```diff
@@ -398,234 +425,73 @@ describe('LaunchForm: Task', () => {
...
+                it('for K8', async () => {
+                    const { getByText, getByLabelText } = renderForm();
+                    await waitFor(() =>
+                        getByLabelText(
+                            AuthRoleStrings[AuthRoleTypes.k8].inputLabel,
+                            {
+                                exact: false
+                            }
+                        )
+                    );
```
csirius, line 452: Same here.

csirius (commented):
Overall LGTM!
I just left some comments for clarification. Please let me know your thoughts. 😀

src/components/Launch/LaunchForm/constants.ts
jsonporter: Whoops, meant to remove this; good catch - fixed.

src/components/Launch/LaunchForm/LaunchRoleInput.tsx
jsonporter: Fixed.

src/components/Launch/LaunchForm/test/LaunchTaskForm.test.tsx
jsonporter: If you look the expect() is on the next line.  We need the async bit because these elements are rendered dynamically so we wait for the element to exist then we query it.

csirius (approved):
LGTM!

PR #254, opened 2021-12-16
feat: show details panel for failed node

Pianist038801:

View Input & task definition of every register version directly

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Tracking Issue

fixes https://github.com/flyteorg/flyte/issues/1697

## Follow-up issue
_NA_


csirius (dismissed):
LGTM!

src/components/Executions/ExecutionDetails/NodeExecutionDetailsPanelContent.tsx
```diff
@@ -218,13 +238,62 @@ const NodeExecutionTabs: React.FC<{
...
 
+const WorkflowTabs: React.FC<{
+    workflow: Workflow;
+    nodeId: string;
+}> = ({ workflow, nodeId }) => {
+    const styles = useStyles();
+    const tabState = useTabState(tabIds, tabIds.inputs);
+    const commonStyles = useCommonStyles();
+    let tabContent: JSX.Element | null = null;
+    const currentNode = workflow.closure?.compiledWorkflow.primary.template.nodes.find(
```
jsonporter, line 249: This will only work for top-level nodes; you need to also collect nested nodes (Subworkflows, branches, dynamic, etc).

assetsTransformer.js
```diff
@@ -0,0 +1,9 @@
+const path = require('path');
+
+module.exports = {
+    process(src, filename, config, options) {
```
anrusina, line 4: nit: IF you will add underscore to function parameter's names that are not in use - > eslint will not warn about them. (in case if you want to leave them named)
in this case: `process(src, filename, _config, _options) {`

src/components/WorkflowGraph/utils.ts
```diff
@@ -116,3 +116,33 @@ export const getTaskTypeFromCompiledNode = (
...
+
+    if (value.taskNode) {
+        return value.taskNode.referenceId.name;
+    } else if (value.workflowNode) {
+        const {
+            workflowNode: {
+                subWorkflowRef: { name }
+            }
+        } = value;
+        return name;
```
anrusina, line 132:
```suggestion      
        return value.workflowNode.subWorkflowRef.name;
```

src/components/Executions/ExecutionDetails/NodeExecutionDetailsPanelContent.tsx
```diff
@@ -238,6 +308,26 @@ export const NodeExecutionDetailsPanelContent: React.FC<NodeExecutionDetailsProp
...
+        );
+        const workflowData: Workflow = await fetchWorkflow(
+            queryClient,
+            workflowExecution.closure.workflowId
+        );
+        if (workflowData) {
+            const keyedDag = transformWorkflowToKeyedDag(workflowData);
+            setDag(keyedDag);
+        }
+    };
```
anrusina, line 324: This is async function, which means it could execute `setDag` after component was already unmounted -> which would lead to a memory leak on the site (see article [here](https://dev.to/alexandrudanpop/correctly-handling-async-await-in-react-components-4h74))

This is pattern I have successfully used so far:
```
useEffect(()=>{
  let isCurrent = true;
  getWorkflowDag().then((keyedDag)=>{
    if (isCurrent) {
      setDag(keyedDag);
    }
  })
  // This handles the unmount case
  return ()=>{ isCurrent = false; }; 
}, [nodeExecutionId])
```

src/components/WorkflowGraph/utils.ts
```diff
@@ -116,3 +116,33 @@ export const getTaskTypeFromCompiledNode = (
...
+    }
+    return '';
+};
+
+export const transformWorkflowToKeyedDag = (workflow: Workflow) => {
+    if (!workflow.closure?.compiledWorkflow) return {};
+
+    const dagData = transformerWorkflowToDAG(
```
anrusina, line 140: nit: Dag (sentence case - similar to other places, where you used Dag in function name)

assetsTransformer.js
Pianist038801: Pushed the fix.

src/components/WorkflowGraph/utils.ts
Pianist038801: Pushed the fix.

src/components/Executions/ExecutionDetails/NodeExecutionDetailsPanelContent.tsx
Pianist038801: Pushed the fix.

src/components/WorkflowGraph/utils.ts
Pianist038801: Pushed the fix.

anrusina (approved):

assetsTransformer.js
```diff
@@ -0,0 +1,9 @@
+const path = require('path');
```
anrusina, line 1: ⚠️ Note: This and file below are also present in David's PR, which means that they will create a merge conflict for each other.

anrusina (approved):

PR #264, opened 2022-01-13
fix: recent workflow bug on filtering executions

Pianist038801:
[BUG-2065](https://github.com/flyteorg/flyte/issues/2065)

This fixes a bug in the recent workflow versions table when filtering executions.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

### DIFF
--- a/src/components/Entities/EntityVersions.tsx
+++ b/src/components/Entities/EntityVersions.tsx
@@ -47,7 +47,6 @@ export const EntityVersions: React.FC<EntityVersionsProps> = ({
 }) => {
     const { domain, project, resourceType, name } = id;
     const styles = useStyles();
-    const filtersState = useWorkflowExecutionFiltersState();
     const sort = {
         key: executionSortFields.createdAt,
         direction: SortDirection.DESCENDING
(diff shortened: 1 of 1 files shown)


src/components/Entities/EntityVersions.tsx
```diff
@@ -62,7 +62,7 @@ export const EntityVersions: React.FC<EntityVersionsProps> = ({
         { domain, project },
         {
             sort,
-            filter: [...baseFilters, ...filtersState.appliedFilters],
+            filter: baseFilters,
```
anrusina, line 65: I see there is three filters which are using `[...baseFilters, ...filtersState.appliedFilters],` currently in code. Can you explain why filterState.appliedFilters is not needed here?
Also if it is removed, than it could also be removed from line 50 and line 82 in this functions, as they become redundant.

anrusina: If the filter is not needed than filtersState mentions in line 50 and line 82 can also be removed 🤔

anrusina (approved):
Smoke-tested - works as expected.
@Professional0321 Can you please  removed unneeded mentions of `filtersState` in that file.

src/components/Entities/EntityVersions.tsx
Pianist038801: @anrusina That's a good catch. Just pushed fix!

anrusina (approved):

Issue #292, opened 2022-03-02
[BUG] In Flyte UI workflow execution name not on the correct place


jganbat-fn:

anrusina:
Node-ID should be wrapped in case of long identifier string. Full string should be shown on hover. Will leave this issue to track that work.

PR #308, opened 2022-03-07
fix: long node-id obscures task type #patch

anrusina:
Wrap text with ellipsis if it's too long, and add Tooltip, so we can still see whole display-id.
**Most changes a whitespace only**, I marked the fix with a comment

Prior to the fix:
Screen Shot 2022-03-07 at 12 30 41 PM

After the fix:
Screen Shot 2022-03-07 at 12 42 03 PM

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [ ] Unit tests added

## Complete description
The main point described in issue 292 was fixed in flyteconsole v.0.42.1
The only issue left - was nodeId is not fully readable and obscuring the view, in case if id is too long to fit into a cell. This PR fixes it.

fixes https://github.com/flyteorg/flyteconsole/issues/292

### DIFF
--- a/src/components/Executions/Tables/nodeExecutionColumns.tsx
+++ b/src/components/Executions/Tables/nodeExecutionColumns.tsx
@@ -1,9 +1,5 @@
-import { Typography } from '@material-ui/core';
-import {
-    formatDateLocalTimezone,
-    formatDateUTC,
-    millisecondsToHMS
-} from 'common/formatters';
+import { Tooltip, Typography } from '@material-ui/core';
+import { formatDateLocalTimezone, formatDateUTC, millisecondsToHMS } from 'common/formatters';
 import { timestampToDate } from 'common/utils';
 import { useCommonStyles } from 'components/common/styles';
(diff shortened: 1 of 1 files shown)


src/components/Executions/Tables/nodeExecutionColumns.tsx
```diff
@@ -17,226 +13,197 @@ import { NodeExecutionCacheStatus } from '../NodeExecutionCacheStatus';
...
+    };
+  });
 
-    return <>{displayId ?? execution.id.nodeId}</>;
+  const nodeId = displayId ?? execution.id.nodeId;
+  return (
+    <Tooltip arrow title={nodeId} placement="top-start">
+      <div className={commonStyles.truncateText}>{nodeId}</div>
+    </Tooltip>
+  );
```
anrusina, line 87: Fix related change

jsonporter (approved):
lgtm

PR #313, opened 2022-03-08
ci: move to webpack 5

anrusina:
This update moves our build to use webpack 5, it is also moves storybook to use webpack 5.

Unfortunately I was needed to avoid CDN link usage for react and react-dom, which increased the size of a built application by 10 Mb. However, it is more important to move to webpack 5 now then to ensure small size of the package. We can spend more time on optimization after we will update our repo to latest packages across the board.

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [ ] Unit tests added


jsonporter (approved):
lgmt

jsonporter (approved):
lgtm

PR #328, opened 2022-03-15
chore: update eslint and prettier, use arbnb rules as base for our new standart  #patch

anrusina:
This PRs updates our **eslint** rules and **prettier** to latest standard.

From code perspective this PR contains only:
- space changes
- minimal import shuffling
- adding trailing comma

I did my best to avoid any code changes, by turning off all eslint rules which we are currently violating.
I have enabled few rules as warn, so we can start working on them. All rules that were enabled as warn can not be autofixed at this point (to avoid automatic unverified fix during commit).

### NOTE
After this PR would be merged, you would want to  pull it locally ASAP, to minimize **diff** amount in future PRs, 

## Type
 - [X] Maintenance
 
## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [X] Unit tests checked


jsonporter (approved):
lgtm

PR #344, opened 2022-03-23
chore: cleaning up eslint rules

anrusina:

Updated some eslint rules settings, to ensure more accurate erroring. 
Fixed errors which were not required more detailed testing:

- Updated `no-unused-vars` and `no-redeclare` to typescript versions
- Allowed to break `camelcase` and `jsx-a11y/*` rules in **test** and **storybook** files
- removed some unused variables, which was obviously not in use (left in place those which require some investigation)
- removed unused exports for Props interfaces (we should limit export usage only if type is needed to be reused - it minimizes release package size)
- moved render functions to the end of class ( in two places) - no code change
- disabled `no-redeclare` rule for enum declarations

## Type
 - [X] Chore
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?
 - [X] Code completed
 - [X] Smoke tested
 - [ ] Unit tests added

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -22,7 +22,10 @@ module.exports = {
     },
   },
   rules: {
-    // "off" or 0: turn off the rule completely; "warn" or 1;  "error" or 2
+    /**
+     * Rules we don't want to be enabled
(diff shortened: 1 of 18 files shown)


olga-union (approved):
LGTM

PR #349, opened 2022-03-30
chore: eslint cleanup - eqeqeq, unused experession, index key

anrusina:
Eslint related cleanup with no logic changes. 
All non-obvious fixes where skipped for now. Covered next rules:

- `eqeqeq` - When == used instead of ===  
-  `react/no-array-index-key` - it's not safe to use index as key in the arrays (it's default behaviour, so don't even have much sense) - replaced where a different unique identified already available
-  removed unused imports and some unused variables
 

## Type
 - [X] Chore
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [X] Smoke tested
 - [ ] Unit tests added

### DIFF
--- a/src/components/Navigation/UserInformation.tsx
+++ b/src/components/Navigation/UserInformation.tsx
@@ -31,7 +31,7 @@ export const UserInformation: React.FC<{}> = () => {
   return (
     <WaitForData spinnerVariant="none" {...profile}>
       <div className={useStyles().container}>
-        {profile.value == null ? (
+        {!profile.value ? (
           <LoginLink />
         ) : !profile.value.preferredUsername || profile.value.preferredUsername === '' ? (
           profile.value.name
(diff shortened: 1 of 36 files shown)


src/components/Navigation/UserInformation.tsx
```diff
@@ -33,7 +33,7 @@ export const UserInformation: React.FC<{}> = () => {
       <div className={useStyles().container}>
         {profile.value == null ? (
```
olga-union, line 34: nit: use `!!` to remove a `==` check?
```suggestion
        {!!profile.value ? (
```

olga-union: oops 😅 yeah, that's what I meant!

olga-union (approved):
LGTM!

Issue #318, opened 2022-03-10
[BUG] On resize Header and body of the tables are mismatched


anrusina:
When you resize the window table's header and columns are resized differently

<img width="850" alt="Screen Shot 2022-03-09 at 5 55 10 PM" src="https://user-images.githubusercontent.com/55718143/157573273-b50dc80b-44b9-461d-9f26-14ca33a92b1f.png">

#### Can be seen in next views
Workflow execution view:
https://demo.nuclyde.io/console/projects/flytesnacks/domains/development/workflows/athena.athena.no_io_wf?duration=all

Nodes view:
https://demo.nuclyde.io/console/projects/flytesnacks/domains/development/executions/gcg0wvmhfh?duration=all

Task execution view:
https://demo.nuclyde.io/console/projects/flytesnacks/domains/development/tasks/athena.athena.print_athena_schema?duration=all

Executions view:
https://demo.nuclyde.io/console/projects/flytesnacks/executions?duration=all

eugenejahn:
when resizing windows to smaller heights, the rows will disappear. The row disappears because the components above already take all the windows height, so the table rows container height set to 0px. 
 
![image](https://user-images.githubusercontent.com/25038146/162598529-699a73a0-7bdf-4a44-b10f-7e1b7e79d64f.png)


![image](https://user-images.githubusercontent.com/25038146/162598533-02c14dec-79ad-4e14-9c29-d261fadaf717.png)

PR #369, opened 2022-04-10
fix: on resize Header and body of the tables are mismatched

eugenejahn:
Signed-off-by: Eugene Jahn <eugenejahnjahn@gmail.com>

fixed this issue https://github.com/flyteorg/flyteconsole/issues/318
btw. I have updated the issue to address another responsive bug for the table. It's also fixed. 

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
removed `flex-grow: 0` and `flex-shrink:0` on the table column. We don't need it because we want the table to be responsive. Otherwise, can't show data in the smaller width window. 

also, set a minHeight for the table rows container. Otherwise, the height of the row container will be set to 0 due to auto resizer.  

## Tracking Issue

fixes: https://github.com/flyteorg/flyteconsole/issues/318

## Follow-up issue
_NA_

### DIFF
--- a/src/components/Executions/Tables/styles.ts
+++ b/src/components/Executions/Tables/styles.ts
@@ -105,8 +105,6 @@ export const useExecutionTableStyles = makeStyles((theme: Theme) => ({
     flexDirection: 'row',
   },
   rowColumn: {
-    flexGrow: 0,
-    flexShrink: 0,
     marginRight: theme.spacing(1),
     minWidth: 0,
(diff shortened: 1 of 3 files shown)


anrusina (approved):
LGTM, performed smoke test. 
Thank you a lot for contributing 🥳🙃

Issue #337, opened 2022-03-21
[INFRA] Add more descriptive errors on task/workflow run failure


anrusina:
### Issues

The error showed in UI is not descriptive enough - in case when task/workflow failed due to UI settings.
For example:  If wrong dimensionality was selected - the errors looks like these:
workflow: `Request failed with status code 400`
task: `Request failed with status code 500`

### What should we show

The error execution respond for [demo task](https://demo.nuclyde.io/console/projects/flytesnacks/domains/development/tasks/core.basic.folders.rotate_all?duration=all) was:
```
¢Collected Errors: 2
	Error 0: Code: MismatchingTypes, Node Id: start-node, Description: Variable [img_dir] (type [blob:<dimensionality:MULTIPART > ]) doesn't match expected type [blob:<> ].
	Error 1: Code: ParameterNotBound, Node Id: start-node, Description: Parameter not bound [img_dir].

¢Collected Errors: 2
	Error 0: Code: MismatchingTypes, Node Id: start-node, Description: Variable [img_dir] (type [blob:<dimensionality:MULTIPART > ]) doesn't match expected type [blob:<> ].
	Error 1: Code: ParameterNotBound, Node Id: start-node, Description: Parameter not bound [img_dir].
```

Issue #336, opened 2022-03-21
[BUG] Dimensionality - should be set per data type blob and be disabled in UI.


anrusina:
### Issue:
FlyteUI dimensionality defaults to Single when the input type is a FlyteDirectory .

### Repro:
Open folder task in DEMO site: 
 - https://demo.nuclyde.io/console/projects/flytesnacks/domains/development/tasks/core.basic.folders.rotate_all?duration=all
- Click "Launch Task" button
- Observe that `img_dir (file/blob)*` expects to get directory as an input, but dimensionality is set to "Single (File)"

[Task template](https://demo.nuclyde.io/api/v1/tasks/flytesnacks/development/core.basic.folders.rotate_all/773447b298bfa8ecfc2b25983ce1ed2d33753d01) sets dimensionality for this data to MULTIPART 
```
"img_dir": {
  "type": {
     "blob": {
        "dimensionality": "MULTIPART"
      }
    },
  "description": "img_dir"
}
``` 

### How should fix look like

Dimensionality should match data contract specified in data type. Control should be disabled, so the user can see type but won't be able to switch it.

### Notify on thread fix:
https://flyte-org.slack.com/archives/CREL4QVAQ/p1646872001275689

eugenejahn:
hi! 
I was trying to solve this issue, can produce it, and have looked into the codebase, but not pretty sure what is the expected behavior. 

It looks like we already matching the dimensionality from the data contract, and if empty, the default dimensionality is set to 0 
https://github.com/flyteorg/flyteconsole/blob/4721e84e43dee81b52f1e89cc1e555529717cb15/src/components/Launch/LaunchForm/BlobInput.tsx#L41

so the only issue I see is that we didn't disable the drop-down. Then we should set the select componet to disabled,  so the user can see type but won't be able to switch it.
https://github.com/flyteorg/flyteconsole/blob/4721e84e43dee81b52f1e89cc1e555529717cb15/src/components/Launch/LaunchForm/BlobInput.tsx#L100-L109


Let me know if that's correct! Thanks. Also, if you want, you can also slack me for communication:)

anrusina:
To get to BlobInput we are first checking `input.typeDefinition.type` in https://github.com/flyteorg/flyteconsole/blob/4721e84e43dee81b52f1e89cc1e555529717cb15/src/components/Launch/LaunchForm/LaunchFormInputs.tsx#L19

[...]

PR #370, opened 2022-04-11
fix: add more descriptive errors on task/workflow run failure

eugenejahn:

Not sure should I suppose to fix this...  I was trying to fix  https://github.com/flyteorg/flyteconsole/issues/336, but kind of annoyed by not seeing the error message, so I just fix it. (may not be 100% complete)

 
Enrich the error message. In launch form, users should see a complete error message from the API response 
After fixed: 
image

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

I decoded the response from the api and return it as new error. However, one thing I am not quite sure is do we have a decode type for the error message? now, I am just using RawOutputDataConfig type to decode...

also since I don't think we will use decodeErrorResponseMessage in other place, I didn't put in ./utils. Let me know if you want me to put into utils. 

## Tracking Issue

fixes https://github.com/flyteorg/flyteconsole/issues/337

### DIFF
--- a/src/models/AdminEntity/transformRequestError.ts
+++ b/src/models/AdminEntity/transformRequestError.ts
@@ -1,9 +1,32 @@
 import { AxiosError } from 'axios';
 import { NotAuthorizedError, NotFoundError } from 'errors/fetchErrors';
+import { Admin } from 'flyteidl';
+import { decodeProtoResponse } from './utils';
+
+// to enrich the error message by adding the response
+function decodeErrorResponseMessage(error: AxiosError) {
+  try {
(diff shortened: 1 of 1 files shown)


anrusina (approved):
It looks great! 🚀
Currently we do not have a decode type for such errors, so what you did is probably the best that can be achieved.
 I fully agree with you that it is a best place to put it in, as it planned to be used with transformRequestError only at this point.

Issue #372, opened 2022-04-12
Bug: Service account is not properly propagated from the execution info


anrusina:
Currently The Service account value is not connected to the value provided in execution description.
We need to fix it and to check that all other values in header are read from proper places.

Example:
https://demo.nuclyde.io/console/projects/flytesnacks/domains/development/executions/apmwh7fspl5wrhx9tx97?duration=all
The execution has set to default:
<img width="1257" alt="Screen Shot 2022-04-12 at 10 20 11 AM" src="https://user-images.githubusercontent.com/55718143/163019199-cf0db46a-7836-4c27-bda6-2bfa82c082f7.png">

However, if we will look into API request:
https://demo.nuclyde.io/api/v1/executions/flytesnacks/development/apmwh7fspl5wrhx9tx97
The Security Context there is set to "Demo":
```
{
  "id": {
      "project": "flytesnacks",
      "domain": "development",
      "name": "apmwh7fspl5wrhx9tx97"
  },
  "spec": {
      "launch_plan": {...}, 
      "inputs": {...},
      "metadata": {},
      "labels": {},
      "annotations": {},
      "security_context": {
          "run_as": {
               "k8s_service_account": "demo"
           }
      }
   },
  "closure": {...}
}
```

anrusina:
@eugenejahn We need to use `security_context` instead of `auth_role` for both **IAM role** and **service Account** values. Looks like security context was added quite a time ago and considered a proper way to check for sometime now.
Also `auth_role` is deprecated in backend and would be fully removed from usage soon, so no need to keep it's support in here.

PR #374, opened 2022-04-13
fix: service account is not properly propagated from the execution info

eugenejahn:

Service account is not properly propagated from the execution info

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
should fix the issue, but can't reproduce the demo bug. so not 100% sure the results. let me know if the code is not working 

## Tracking Issue

fixes https://github.com/flyteorg/flyteconsole/issues/372

## Follow-up issue
_NA_
OR

### DIFF
--- a/src/components/Executions/ExecutionDetails/ExecutionMetadataExtra.tsx
+++ b/src/components/Executions/ExecutionDetails/ExecutionMetadataExtra.tsx
@@ -38,7 +38,6 @@ export const ExecutionMetadataExtra: React.FC<{
     launchPlan: launchPlanId,
     maxParallelism,
     rawOutputDataConfig,
-    authRole,
     securityContext,
   } = execution.spec;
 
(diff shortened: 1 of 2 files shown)


src/components/Executions/ExecutionDetails/ExecutionMetadataExtra.tsx
```diff
@@ -54,7 +55,10 @@ export const ExecutionMetadataExtra: React.FC<{
     },
     {
       label: ExecutionMetadataLabels.serviceAccount,
-      value: authRole?.kubernetesServiceAccount || ExecutionMetadataLabels.securityContextDefault,
+      value:
+        authRole?.kubernetesServiceAccount ||
```
anrusina, line 59: Just saw you comment in the task, we don't need to support for `authRole` any more as it's a deprecated field, so line 59 should be safely removed.

anrusina: Can you please also update line 53 to use `securityContext.IamRole` instead of `authRole`

anrusina: Also this lines could also now be safely removed: https://github.com/flyteorg/flyteconsole/blob/24bdaeefcf32c079d1aaeb54b28cc6bf7395b42f/src/models/Execution/api.ts#L125-L127
😇

eugenejahn:
Sounds good. Updated!

anrusina (approved):
LGTM, smoke tested.

Thank you for helping us 
# 🥳

Issue #378, opened 2022-04-13
[Graph] Some graphs are not rendering due to Cannot read properties of undefined (reading 'id')


anrusina:
### Issue

For some corner case executions Graphs are not rendered due to "Cannot read properties of undefined (reading 'id')" issue.
Example: 
https://demo.nuclyde.io/console/projects/flytesnacks/domains/development/executions/f9487ac05ff2e4a03931?duration=all&fromExecutionNav=true

<img width="605" alt="Screen Shot 2022-04-13 at 12 55 30 PM" src="https://user-images.githubusercontent.com/55718143/163263909-47dacb53-ea95-4ebc-94ba-83e9845155a6.png">

### Goal

The Graph should be shown. 
Taking in mind that timeline is rendered properly, it looks like some null check is missing

Issue #377, opened 2022-04-13
[Graph] Graph view need to support Branch nodes


anrusina:
### Issue 

"Currently branch node view is messed up and hard to follow" - per Ketan statement.
Examples: 
`control_flow.run_merge_sort.merge_sort` - only failed items are rendered.
https://demo.nuclyde.io/console/projects/flytesnacks/domains/development/executions/f1c3d1514116a44019fe?duration=all
`versions view`
https://demo.nuclyde.io/console/projects/flytesnacks/domains/development/workflows/core.control_flow.run_merge_sort.merge_sort/version/ARW5epeOV80feOjLTUXmtQ==
`dolt.dolt_branch_example.wf`
https://demo.nuclyde.io/console/projects/flytesnacks/domains/development/executions/arnb8thjdrpqqmsgqrmc?duration=all

### Solution

At this point we haven't treated the branch node in any way that is different from regular python_task in DagV2, as such we don't have enough information to show it in a proper way

This work should be approached in few steps:
[ ] Understand Branch node type (spike)
[ ] Design requirements for how we will show branch node in Graph View
[ ] Add Branch node to DagV2 info (?)

cosmicBboy:
won't fix

PR #388, opened 2022-04-14
fix(378): ensure that undefined executions ids won't get into graph

anrusina:
Current fix removed request for undefined executionIds, which usually happens with branch nodes.
It allows us to avoid unneeded crash. Further work is needed for proper support of Branch nodes.

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [ ] Unit tests added

## Tracking Issue

fixes https://github.com/flyteorg/flyteconsole/issues/378

## Follow-up issue
This fix currently removes the crash, however it doesn't fix the branch node representation, which is still a bit off. A separate task was created to cover that work: 
https://github.com/flyteorg/flyteconsole/issues/377
`TODO` with task number was added to the code to ensure, that fix would be fully tested

### DIFF
--- a/src/components/Workflow/workflowQueries.ts
+++ b/src/components/Workflow/workflowQueries.ts
@@ -1,3 +1,4 @@
+import { log } from 'common/log';
 import { QueryInput, QueryType } from 'components/data/types';
 import { extractTaskTemplates } from 'components/hooks/utils';
 import { getNodeExecutionData } from 'models/Execution/api';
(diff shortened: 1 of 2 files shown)


src/components/Workflow/workflowQueries.ts
```diff
@@ -13,9 +14,10 @@ export function makeWorkflowQuery(queryClient: QueryClient, id: WorkflowId): Que
       // On successful workflow fetch, extract and cache all task templates
       // stored on the workflow so that we don't need to fetch them separately
       // if future queries reference them.
-      extractTaskTemplates(workflow).forEach((task) =>
-        queryClient.setQueryData([QueryType.TaskTemplate, task.id], task),
-      );
+      extractTaskTemplates(workflow).forEach((task) => {
+        return queryClient.setQueryData([QueryType.TaskTemplate, task.id], task);
```
olga-union, line 18: What's the purpose of the `return` here?

olga-union, line 18: What's the purpose of the `return` here?

anrusina: Will remove, was adding `{}` curely braces during debug. Curly braces as not needed anymore and as such return not needed too. 
Will update in a miunte

olga-union (approved):
LGTM!

Issue #324, opened 2022-03-10
feat: "Only Mine" Global Toggle


anrusina:
### Motivation: Why do you think this is important?

Currently users have no way to filter execution lists by only their executions.  To fix this we will add an "Only Mine" toggle in Main Application Bar

### Goal: What should the final outcome look like, ideally?

All lists that have filters will check the state of the "Only Mine" filter.  Additionally, all lists will include a subtext that indicates they are showing results filtered by "Only Mine" as a tag.  Clicking to close that tag will also set the toggle button to disabled state.

By default toggle button will enable "Only me" view for all sections. User will be able to update toggle setting to include only views they are interested in.

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

anrusina:
Additional info for our criteria after conversation with [Eugene](https://github.com/eugenejahn):
The “The user selections are the source of truth” criteria, refers to an additional toggle we currently have in executions list (Project Dashboard and Workflow ->individual pages) - for **only my executions.** We will have two points to control only my visibility for “Executions”:
- Upper “Only my” toggle
- Checkbox on the page.

They will merge statuses in next manner:

1. When you enter the page CheckBox status should read value from Toggle settings, show it and apply it (as it does currently) to the query
2. If you are updating the CheckBox status by clicking on it -> it will NOT update toggle selection (Main point of truth) but will start apply new filter to the query on the page. It means that if you leave this page and than redirect back - checkbox status will be again read from toggle

anrusina:
'Only my launch form' need to be replaced with 'Only my launch plans'
It will be connected in two places:

- one in LaunchWorkflow -> Launch plan dropdown
- in Launch plan list page -> After it would be implemented #298

anrusina:
@pmahindrakar-oss 
We need confirmation if next APIs can be filtered by user id (ideally with sample of API call), and if not - is it possible to do at all (missing API):
[List of workflows](https://demo.nuclyde.io/api/v1/named_entities/2/flytesnacks/development?filters=eq(state,0)&token=&limit=4)  - looks like it’s not supported. Also looks like this one also returns one less item then specified in limit
[List of tasks](https://demo.nuclyde.io/api/v1/named_entities/1/flytesnacks/staging?token=&limit=4)
[...]

PR #411, opened 2022-04-19
chore: preparations for "Only Mine" feature support

eugenejahn:
# TL;DR
still working on fixing the styles and reformating the code, so plz ignore the code change 
send this PR to ensure the logic is correct. hope someone can verify with a smoke test. 

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
- saved the only mine setting and toggle in the local storage
- have a context provider in order to pass local storage to children component 
- still working on cleaning the format and add some comments 

## Tracking Issue

Work on: https://github.com/flyteorg/flyteconsole/issues/324


src/components/Navigation/OnlyMine/constant.ts
```diff
@@ -0,0 +1,37 @@
+export const filterByDefault = [
```
anrusina: **nit:** // usually comment mark as "**nit:**" at the beginning - can be treated as optional change
I would prefer the default setting to leave in `src/basics/LocalCache/defaultConfig.ts` or be in a separate file located in `src/basics/LocalCache/`. In this case I would suggest to use filter as 
```
{
selectAll : true,
onlyMyWorkflow: true, 
...
}
```
In this case labels will leave in a separate string files in OnlyMine folder. String file will look similar to: `src/components/Executions/Tables/WorkflowExecutionTable/strings.ts` and can be used as `t('inputOutputTooltip')`

We just starting moving to string files. It's a good practice to put all your strings in a separate locations, as it's simplifies localization in future.

This change is fully optional, as the pattern you used is all over out code at this point.

src/basics/LocalCache/ContextProvider.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+export const LocalCacheContext = createContext<LocalCacheState>({
+  localCacheMap: { ...allStorage() },
+  setLocalCache: () => {
+    /* Provider is not initialized */
+  },
+  getLocalCache: () => false,
+});
+
+export const LocalCacheProvider = (props: LocalCacheProviderProps) => {
```
anrusina, line 41: I'm curios why localCacheProvider was needed, did you get to many rerenders?

anrusina (commented):
Looks pretty good! 
 left a set of comments, mainly to align with our evolving practices, as there is not enough code samples.

From UI perspective there is two notes:
- The "Reset" button should be all the way to the right in popover
- Not only Hamburger but also text "PersonalMOde" should open popover.  You can wrap it with div and add click  handler on it. ensure that div has aria-role set to button and aria-lable set to something like "Open Popover to select personal filter prefernces". Aria labels needed for accessibility
- There is also some glitch which I was able to see without 100% repo (4/10 times) when we get FailedToFetchData. I see that in those cases we tried to filter providing empty userID as `https://demo.nuclyde.io/api/v1/executions/flytesnacks/development?filters=eq(workflow.name,aa.wine_study.wf)**%2Beq(user,)**&token=&limit=100&sort_by.direction=DESCENDING&sort_by.key=created_at`
Looks like there could be a race condition where UserId inofrmation comes after the query already finished. We can sync on it separately if needed.

src/basics/LocalCache/ContextProvider.tsx
eugenejahn: since we want to share the Only Mine Setting State between NavBar and the checkmark in OnlyMyExecuations and other components. Therefore, we need to use a provider to wrap all children's components. Let me know if that doesn't make sense~

anrusina: `LocalCache` is stored in Browser Local Storage, which means it always has only one instance and always accessible for the code - even if we access it from different parts of the code or between app use.  You can find it here in DevTools:
Screen Shot 2022-04-20 at 2 17 00 PM

With this in mind CacheProvider shouldn't be needed to access the data, as previous useHook will read it from LocalStorage directly each time you request it. So in this case CacheProvider could be needed only in case if you think that we would have to many read/write operations so it will affect the performance. However, the reading/writing to LocalStorage from my experience is so inexpensive, that it doesn't really need to have a wrapper.

anrusina: Adding the Provider here - can even introduce issues, as now you additionally cache the value in the app. So if I will go to local storage and update the value, it won't be picked up on re-navigation.

anrusina: Oh I see why you did it. You needed a timed update, when you changed it one place, to notify the page that it was changed too.. I will think if we need to do anything around it, as I would like to avoid a Provider for all LocalStorage items here.

eugenejahn: yea. I need to get the update when OnlyMine changes the setting values. otherwise, can't update the checkbox in the dashboard. I see why you want to avoid a Provider for all LocalStorage items here. either we can find a solution to listen the localstorage change and update the state in Context or maybe we should create a provider just for navbar or general setting? 

(btw. normally I used redux for managing local storage, and redux will treat local storage as the normal state with the hook. so it won't have this issue..)

eugenejahn: btw. i think you change the value in localstorage and refresh. it should update for now.

anrusina: Yes, let's stick with provider. If you updated it manually it should be okay to require page refresh. 
You will need to update unit tests for it.
You can run tests per test file like that:
`yarn test LocalCache`

src/components/Theme/muiTheme.ts
```diff
@@ -176,5 +176,27 @@ export const muiTheme = {
...
+        },
+      },
+      track: {
+        // Controls default (unchecked) color for the track
+        opacity: 1,
+        backgroundColor: '#E6E7E8',
+        '$checked$checked + &': {
+          // Controls checked color for the track
+          opacity: 1,
+          backgroundColor: '#73BE6D',
```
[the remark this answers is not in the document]
eugenejahn: fixed

src/components/common/Icons/DropdownIcon.tsx
```diff
@@ -0,0 +1,40 @@
+import * as React from 'react';
+import { IconProps } from './interface';
+
+export const DropdownIcon: React.FC<IconProps> = ({ size = 17, className, onClick }) => {
```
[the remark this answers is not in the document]
eugenejahn: replaced with MUI icom

src/components/Navigation/OnlyMine/constant.ts
eugenejahn: moved the config to localcache folder, and added the string file in onlymine folder

.storybook/StorybookContainer.tsx
```diff
@@ -9,7 +9,8 @@ import { SnackbarProvider } from 'notistack';
 import { ErrorBoundary } from '../src/components/common/ErrorBoundary';
 import { createQueryClient } from '../src/components/data/queryCache';
 import { muiTheme } from '../src/components/Theme/muiTheme';
-
+import { LocalCacheProvider } from '../src/basics/LocalCache/ContextProvider';
+import { FeatureFlagsProvider } from '../src/basics/FeatureFlags';
 const useStyles = makeStyles((theme: Theme) => ({
```
anrusina, line 14: nit: new line between imports and const.
The easiest way to solve most of styling issues is to run `yarn fix` on top level of the repo.

src/basics/LocalCache/localCache.test.tsx
```diff
@@ -10,7 +11,7 @@ const TestFrame = () => {
 
   const toShow = () => setShow(true);
   const toClear = () => clearShow();
-
+  console.log('show', show);
```
anrusina, line 14: nit: remove the console.log

src/basics/LocalCache/onlyMineDefaultConfig.ts
```diff
@@ -0,0 +1,47 @@
+export enum OnlyMyFilter {
+  // Test flag is created only for unit-tests
```
anrusina, line 2: it doesn't look like it used only in tests, is this comment is still valid or should be removed

src/components/Navigation/OnlyMine/FilterPopoverIcon.tsx
```diff
@@ -0,0 +1,75 @@
...
+      cursor: 'pointer',
+      '&:hover': {
+        color: fade(interactiveTextColor, 0.4),
+      },
+    },
+    popoverContent: {
+      border: `1px solid ${theme.palette.divider}`,
+      borderRadius: 4,
+      marginTop: theme.spacing(0.25),
+      padding: `${theme.spacing(2)}px ${theme.spacing(1.5)}px`,
```
anrusina, line 23: nit: `padding: theme.spacing(2, 1.5),`
should have the same value

src/components/Navigation/__stories__/Navbar.stories.tsx
```diff
@@ -4,6 +4,7 @@ import { ExecutionDetailsAppBarContent } from 'components/Executions/ExecutionDe
 import { mockExecution } from 'models/Execution/__mocks__/mockWorkflowExecutionsData';
 import { createMockExecution } from 'models/__mocks__/executionsData';
 import * as React from 'react';
+import { FeatureFlag, useFeatureFlagContext, FeatureFlagsProvider } from 'basics/FeatureFlags';
```
anrusina, line 7: nit: FeatureFlagsProvider is not needed here, and can be safely removed.

anrusina (commented):
Looks pretty good, 
most comments I left are nits to better align with our codebase style.

I changed the title to include "chore: " prefix, so it won't trigger a release on merge it to master. This code is plan to be hidden, as we actively discussing  API updates for it. I don't expect to have API ready until second week of May. So it makes sense to merge the change as is (as it is hidden by Feature Flag) and I will update task with information of what still need ti be done (including storybook and tests)

src/components/Navigation/__stories__/Navbar.stories.tsx
eugenejahn: fixed

src/components/Navigation/OnlyMine/FilterPopoverIcon.tsx
eugenejahn: got it fixed

src/basics/LocalCache/onlyMineDefaultConfig.ts
eugenejahn: fixed

src/basics/LocalCache/localCache.test.tsx
eugenejahn: fixed

src/basics/LocalCache/ContextProvider.tsx
```diff
@@ -0,0 +1,67 @@
...
+interface LocalCacheProviderProps {
+  children?: React.ReactNode;
+}
+
+// get the local storage value, if local storage is null, use default value
+function allStorage() {
+  const defaultKeys = Object.keys(defaultLocalCacheConfig);
+
+  var localCacheMap: Record<string, any> = {},
+    i = defaultKeys.length;
```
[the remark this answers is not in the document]
eugenejahn: fixed

src/basics/LocalCache/ContextProvider.tsx
```diff
@@ -0,0 +1,67 @@
+// More info on Local storage: https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage
+import { log } from 'common/log';
+import * as React from 'react';
+import { useState, createContext, useContext, useCallback } from 'react';
```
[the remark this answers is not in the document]
eugenejahn: fixed

.storybook/StorybookContainer.tsx
eugenejahn: fixed (btw. look like yarn fix didn't catch it)

anrusina (approved):
Great job 

# 🚀

PR #417, opened 2022-04-21
Add interruptible override to launch forms #chore

MorpheusXAUT:
# TL;DR
Adds an option to override the `interruptible` flag for a single execution to launch forms.

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [X] Unit tests added

## Complete description
A new `LaunchInterruptibleInput` component has been added, providing a small form input for marking a single execution of a task or workflow as interruptible via a checkbox. Both `LaunchTaskForm` and `LaunchWorkflowForm` now include this additional input - it has been added to the advanced options of a workflow launch form.
Relaunching an execution applies the previously selected interruptible override setting to the new launch form.
Additionally, a simple `true`/`false` indication has been added to the execution metadata overview, displaying whether an execution had its interruptible flag overwritten.

Some very basic tests have been added to check correct application of the checkbox' value. Manual smoke testing for workflow and single task executions has been performed together with flightadmin and flightpropeller.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2284

## Follow-up issue
_NA_


MorpheusXAUT:
Looks like checks failed due to unreleased flyteidl changes. Is there anything that should be done before the IDL changes are merged (like with `replace` in `go.mod`), or should I update my PR once the spec changes are available?

MorpheusXAUT:
@anrusina Unfortunately still WIP due to pending flyteidl merges, but a slight change has been made already. As the `interruptible` flag is optional and supports an override to enable and disable the setting for an execution, I've having to visualize a boolean with 3 states :smile: 
For now, I've come up with the following:

default, no override
default, no override

override to enabled
override to enabled

override to disabled
override to disabled

Clicking the checkbox cycles through `not set` -> `enabled` -> `disabled` -> `not set`, updating the label for additional clarity.

As I'm neither a great frontend developer nor an UX expert, I'm not sure if there's a better way to represent this while being more clear. I could of course add two separate checkboxes/toggles a la "Enable override" and "Override to", however I found that a bit... tedious to use.

Is there a better/preferred way you can think of to visualize this new setting?

anrusina:
> Clicking the checkbox cycles through `not set` -> `enabled` -> `disabled` -> `not set`, updating the label for additional clarity.
> As I'm neither a great frontend developer nor an UX expert, I'm not sure if there's a better way to represent this while being more clear. I could of course add two separate checkboxes/toggles a la "Enable override" and "Override to", however I found that a bit... tedious to use.

@MorpheusXAUT I don't think that two toggles will make it more clear, and solution you come up with is pretty clever. I asked our designer to look into it - to see if he can suggest something better in this case

src/components/Launch/LaunchForm/LaunchInterruptibleInput.tsx
```diff
@@ -0,0 +1,99 @@
...
+import { Protobuf } from 'flyteidl';
+import { useStyles } from './styles';
+import { LaunchInterruptibleInputRef } from './types';
+import { formStrings } from './constants';
+
+const isValueValid = (value: any) => {
+  return value !== undefined && value !== null;
+};
+
+export interface LaunchInterruptibleInputProps {
```
anrusina, line 14: nit: no need for `export` keyword in this inteface, as it used only in this file

src/components/Launch/LaunchForm/LaunchInterruptibleInput.tsx
```diff
@@ -0,0 +1,99 @@
...
+  initialValue?: Protobuf.IBoolValue | null;
+}
+
+export const LaunchInterruptibleInputImpl: React.ForwardRefRenderFunction<
+  LaunchInterruptibleInputRef,
+  LaunchInterruptibleInputProps
+> = (props, ref) => {
+  const [interruptible, setInterruptible] = React.useState(false);
+  const [indeterminate, setIndeterminate] = React.useState(true);
```
anrusina, line 23: nit: I think a small comment what `indeterminate` means/represents here could be helpful for future developer's sake.

anrusina (commented):
Beautiful code 😍, left two nits which could be ignored.
I definitely haven't smoked it yet, as flyteidl changes I believe still need some time to cook..
Package.k=json will need to point to new flyteidl version when it would be ready.

I will check tomorrow with our designer if he had any suggestions  on UI change for you :)

hongxunjin:
> tedious

@MorpheusXAUT (cc @anrusina ) Overall, it is awesome!  One thing that could possibly be improved is to set the color of the text [interruptible(no override)] as grey when the checkbox turns to "not set", which can visually differentiate "not set" and "disabled".

MorpheusXAUT:
> @MorpheusXAUT (cc @anrusina) Overall, it is awesome! One thing that could possibly be improved is to set the color of the text [interruptible(no override)] as grey when the checkbox turns to "not set", which can visually differentiate "not set" and "disabled".

Sounds like a good plan! How do you like the following (just setting `color="textSecondary"`)

indeterminate
indeterminate state

checked
checked state

hongxunjin:
> indeterminate

It looks great!

MorpheusXAUT:
Feedback from the review has been included, I'll push another commit and update this issue once the IDL changes are merged :slightly_smiling_face:

MorpheusXAUT:
@anrusina flyteidl/npm package has been published, PR should build locally/in CI now :partying_face:
Sorry for the delay!

anrusina (approved):
LGTM, some tests are needed, than could be merged.

Issue #414, opened 2022-04-20
[I/O] Add support for StructuredDataSet Input/Output type.


anrusina:
### Motivation

The Flyte Backend have added new type: *StructuredDataSet* which is currently used as a base one for all new workflows and task in flytesnacks. However UI has no support for it currently and displays such inputs/outputs as:
`df: This type is not yet supported.`

### Info 
Old types are using scalar ` Schema`:
 https://demo.nuclyde.io/console/projects/flytelab/domains/development/executions/f064faa047b702f78000?duration=all&fromExecutionNav=true
here is the task with the schema as an output
https://demo.nuclyde.io/console/projects/flytelab/domains/development/tasks/app.workflow.get_weather_data?duration=all

The new ones with scalar `StructuredDataSet` as input /output - it is very similar to  `Schema` scalar.
https://demo.nuclyde.io/console/projects/flytesnacks/domains/development/executions/fbcca9f3d1d7f44eba5a?duration=all&fromExecutionNav=true

Ping Yee or Nastya for more info

PR #430, opened 2022-04-26
fix: add support for StructuredDataSet

ursucarina:
New feature: Add support for StructuredDataSet 

Example: https://share.getcloudapp.com/z8uL9dQ4
Example: https://share.getcloudapp.com/qGu74YO1

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
Added support for StructuredDataSet 

## Tracking Issue
fixes https://github.com/flyteorg/flyteconsole/issues/414

### DIFF
--- a/packages/zapp/console/src/components/Literals/Scalar/StructuredDatasetValue.tsx
+++ b/packages/zapp/console/src/components/Literals/Scalar/StructuredDatasetValue.tsx
@@ -1,0 +1,114 @@
+import {
+  SimpleType,
+  StructuredDataset,
+  StructuredDatasetColumn,
+  StructuredDatasetColumnType,
+} from 'models/Common/types';
+import * as React from 'react';
+import { PrintList } from '../PrintList';
+import { PrintValue } from '../PrintValue';
+import { useLiteralStyles } from '../styles';
+import { ValueLabel } from '../ValueLabel';
+
+function simpleColumnTypeToString(type: SimpleType) {
+  switch (type) {
+    case SimpleType.NONE:
+      return 'none';
+    case SimpleType.INTEGER:
+      return 'integer';
+    case SimpleType.FLOAT:
+      return 'float';
+    case SimpleType.STRING:
+      return 'string';
+    case SimpleType.DATETIME:
+      return 'datetime';
+    case SimpleType.DURATION:
+      return 'duration';
+    case SimpleType.BINARY:
+      return 'binary';
+    case SimpleType.ERROR:
+      return 'error';
+    case SimpleType.STRUCT:
+      return 'struct';
+    default:
+      return 'unknown';
+  }
+}
+
+function columnTypeToString(literalType?: any, level = 0) {
+  if (level === 2) {
+    return literalType?.type;
+  }
+
+  switch (literalType?.type) {
+    case 'collectionType': {
+      return `collection of ${columnTypeToString(literalType.collectionType, ++level)}`;
+    }
+    case 'simple': {
+      return simpleColumnTypeToString(literalType.simple);
+    }
+    case 'schema':
+      return 'schema';
+    case 'mapValueType':
+      return 'mapValue';
+    case 'enumType':
+      return 'enum';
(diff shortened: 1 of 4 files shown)


packages/zapp/console/src/components/Literals/Scalar/StructuredDatasetValue.tsx
```diff
@@ -0,0 +1,114 @@
...
+  if (level === 2) {
+    return literalType?.type;
+  }
+
+  switch (literalType?.type) {
+    case 'collectionType': {
+      return `collection of ${columnTypeToString(literalType.collectionType, ++level)}`;
+    }
+    case 'simple': {
+      return simpleColumnTypeToString(literalType.simple);
```
ursucarina, line 48: @anrusina what do you think of this vs 
```suggestion
      return (simpleTypeToInputType[literalType.simple] as string).toLocaleLowerCase()
```

packages/zapp/console/src/components/Literals/Scalar/test/StructuredDatasetValue.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+              literalType: {
+                type: 'collectionType',
+                collectionType: {
+                  simple: 2,
+                  type: 'simple',
+                } as any,
+              } as any,
+            },
+          ],
+          format: 'parquet',
```
anrusina, line 54: As we sync today `format: 'parquet'` info should also be added to the data

packages/zapp/console/src/components/Literals/Scalar/StructuredDatasetValue.tsx
```diff
@@ -0,0 +1,122 @@
...
+  StructuredDatasetColumn,
+  StructuredDatasetColumnType,
+} from 'models/Common/types';
+import * as React from 'react';
+import { PrintList } from '../PrintList';
+import { PrintValue } from '../PrintValue';
+import { useLiteralStyles } from '../styles';
+import { ValueLabel } from '../ValueLabel';
+
+function getSimpleTypeDisplayValue(type: SimpleType) {
```
anrusina, line 13: nit: Maybe we can implement it similar to 
```
export const taskTypeToNodeExecutionDisplayType: {..}
```
then usage would be `simpleTypeDisplayValue(type) ?? unknown;`

Totally your call, I just not a big fun of such switches.

anrusina: One more way to do it create a string file and use it from there (similar to   tableLabel_name: 'execution id',):
```
const str = {
  simpleType_: 'unknown',
  simpleType_0: 'none',
  simpleType_1: 'integer'
  ...
}
```
and than call the string as ` t(patternKey('simpleType', type))`
I don't think current implementation will resolve unsupported type to simpleType_ , so additional simpleType_unknown string could be added instead, with type ?? 'unknown' check in t() call...

p.S> Not sure this is a good choice for this function, but for one below - it could be a perfect chioce.

packages/zapp/console/src/components/Literals/Scalar/StructuredDatasetValue.tsx
anrusina: I like when it lowercased - it looks nice. From pattern perspective first one looks better (mainly because i'm not a big fan of `as` casting :)

anrusina (approved):
Let me know if you want to merge it as intermediate step.
Otherwise - it would be nice to add this new type to storybook when you'll switch to a new component

packages/zapp/console/src/components/Literals/Scalar/test/StructuredDatasetValue.test.tsx
ursucarina: this has been added

Issue #398, opened 2022-04-18
[NodeExecution] Make whole row clickable to open TaskExecutionDetails panel


anrusina:
### Motivation

Currently in NodeExecution view only TaskName is a clickable link which leads to open TaskDetailsPanel. 
When the screen size became smaller this table column is fully hidden, and user looses ability to open taskDetailsPanel for the item
<img width="600" alt="Screen Shot 2022-04-18 at 9 36 36 AM" src="https://user-images.githubusercontent.com/55718143/163840884-8a88d294-f775-4d66-a2fb-3c4ab8891802.png">
<img width="600" alt="Screen Shot 2022-04-18 at 9 36 27 AM" src="https://user-images.githubusercontent.com/55718143/163840910-f69720bd-7125-44a0-8c4d-83d705be8a93.png">

### Goal

Ensure that whole row is clickable
Ensure that click on a ViewLogs still leads to logs and doesn't open the side panel

PR #444, opened 2022-04-29
Make whole row clickable to open TaskExecutionDetails panel

eugenejahn:
# TL;DR
make the row clickable and will open the sidebar

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

fixeshttps://github.com/flyteorg/flyteconsole/issues/398

## Follow-up issue
_NA_
OR

### DIFF
--- a/packages/zapp/console/src/components/Executions/Tables/utils.ts
+++ b/packages/zapp/console/src/components/Executions/Tables/utils.ts
@@ -4,3 +4,8 @@ import { nameColumnLeftMarginGridWidth } from './styles';
 export function calculateNodeExecutionRowLeftSpacing(level: number, spacing: Spacing) {
   return spacing(nameColumnLeftMarginGridWidth + 3 * level);
 }
+
+export function selectExecution(state, nodeExecution) {
+  // use null in case if there is no execution provied - to close panel
+  state.setSelectedExecution(nodeExecution?.id ?? null);
+}
--- a/packages/zapp/console/src/components/Executions/Tables/NodeExecutionRow.tsx
+++ b/packages/zapp/console/src/components/Executions/Tables/NodeExecutionRow.tsx
@@ -14,7 +14,7 @@ import { ExpandableExecutionError } from './ExpandableExecutionError';
 import { NodeExecutionChildren } from './NodeExecutionChildren';
 import { RowExpander } from './RowExpander';
 import { selectedClassName, useExecutionTableStyles } from './styles';
-import { calculateNodeExecutionRowLeftSpacing } from './utils';
+import { calculateNodeExecutionRowLeftSpacing, selectExecution } from './utils';
 
 interface NodeExecutionRowProps {
   abortMetadata?: Admin.IAbortMetadata;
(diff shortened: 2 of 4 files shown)


eugenejahn:
Goal
...
Ensure that click on a ViewLogs still leads to logs and doesn't open the side panel

---
However, I think the current logic (before the PR) for ViewLogs leads to opening the side panel. not sure is a bug or not.

anrusina:
> However, I think the current logic (before the PR) for ViewLogs leads to opening the side panel. not sure is a bug or not.

Thank you for pointing it out, we will cover it with some other changes. As I don't think we can really open the logs here anyway, as for some tasks there could be multiple logs.

packages/zapp/console/src/components/Executions/Tables/utils.ts
```diff
@@ -4,3 +4,8 @@ import { nameColumnLeftMarginGridWidth } from './styles';
 export function calculateNodeExecutionRowLeftSpacing(level: number, spacing: Spacing) {
   return spacing(nameColumnLeftMarginGridWidth + 3 * level);
 }
+
+export function selectExecution(state, nodeExecution) {
+  // use null in case if there is no execution provied - to close panel
```
anrusina, line 9: nit: provided  (missing one d)

packages/zapp/console/src/components/Executions/Tables/utils.ts
```diff
@@ -4,3 +4,8 @@ import { nameColumnLeftMarginGridWidth } from './styles';
...
+  state.setSelectedExecution(nodeExecution?.id ?? null);
+}
```
anrusina, line 11: I'm not sure it makes sense to add a helper function for such type of call. I think it would be cleaner to just call state.setSelectedExecution(nodeExecution?.id ?? null) in place. Comment about why null is used could leave near `setSelectedExecution` definition if you are trying to avoid it's duplication.

Anyway, if you want to stick with a helper function -> please provide parameters types you are expecting, as:
```
export function selectExecution(state: NodeExecutionsTableState, nodeExecution: NodeExecution) { 
...
}
```

anrusina (commented):
The build check is failing as expected.
Tests check is already properly connected, so you would ensure that that one is green. Looks like with your changes `NodeExecutionsTable` test is failing, to test only it please run tests from console directory as:
```
narusina@nastyas-mbp flyteconsole % cd packages/zapp/console
narusina@nastyas-mbp console % yarn test NodeExecutionsTable
```

packages/zapp/console/src/components/Executions/Tables/utils.ts
eugenejahn: agree with that since the logic is pretty simple.

anrusina (approved):
Nice work,
### 🎉

Issue #431, opened 2022-04-27
[CI] Stabilizing DevMain branch


anrusina:
Current tasks that need to be done

Move to top level run
- [x] Yarn test
- [x] Yarn fix (lint / format)
- [x] Storybook

Dev simplification:
- [x] Ensure that vscode navigates through all relative paths. // no red herring
- [x] Move PR conditions to devmain branch

anrusina:
Storybook related PR: https://github.com/flyteorg/flyteconsole/pull/434
Unifying tsconfig: https://github.com/flyteorg/flyteconsole/pull/436

anrusina:
Pre-ambula:
* VSCode only looks at closest tsconfig file to figureout how and if Intellisence should be enabled
* Webpack can be configured to look to a specifig tsconfig file

Proper solution to enable intellisence, but avoid building tests/mocks/stories into the final package is:

1. Remove test/spec/mock from exlude in tsconfig.base.json.
2. tsconfig.json inside sub-package extend from base config
3. Create tsconfig.build.json - which is based from nearby located tsconfig.json
4. Force webpack /tsc to use tsconfig.build.json

😜 I wasn’t able to force webpack to use custom tsconfig, in few hours of investigation yesterday, theoretically it should go into ts-loader options:
  use: [
    {
      loader: 'ts-loader',
      options: {
        transpileOnly: true,
        configFile: 'tsconfig.build.json', //here
      },
    },
  ],

Will look into it more closer to the end of week, but if anyone wants to play with it more - feel free :slightly_smiling_face:

PR #451, opened 2022-05-05
chore(tsc): exclude test/stories/mock files only from build

anrusina:
This PR separates default config used from build config to:
- ensure that VSCode IntelliSence will properly map files
- builds will still exclude test / stories / mock files.

In this PR I checked and fixed "tsc" builds - which is used for plugins (and will be used for most other packages)

It DOES NOT include proper build config usage for webpack, yet. It shouould be done separately

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [ ] Unit tests added

## Tracking Issue

part of https://github.com/flyteorg/flyteconsole/issues/431

### DIFF
--- a/packages/plugins/components/package.json
+++ b/packages/plugins/components/package.json
@@ -13,8 +13,8 @@
     },
     "scripts": {
         "build": "yarn build:esm && yarn build:cjs",
-        "build:esm": "tsc --module esnext --outDir lib/esm -p ./tsconfig.build.json",
-        "build:cjs": "tsc -p ./tsconfig.build.json",
+        "build:esm": "tsc --module esnext --outDir lib/esm --project ./tsconfig.build.json",
+        "build:cjs": "tsc --project ./tsconfig.build.json",
         "test": "NODE_ENV=test jest"
     },
     "dependencies": {
--- a/packages/plugins/components/package.json
+++ b/packages/plugins/components/package.json
@@ -13,8 +13,8 @@
     },
     "scripts": {
         "build": "yarn build:esm && yarn build:cjs",
-        "build:esm": "tsc --module esnext --outDir lib/esm",
-        "build:cjs": "tsc",
+        "build:esm": "tsc --module esnext --outDir lib/esm -p ./tsconfig.build.json",
+        "build:cjs": "tsc -p ./tsconfig.build.json",
         "test": "NODE_ENV=test jest"
     },
     "dependencies": {
--- a/packages/plugins/components/src/NavBar/navbar.stories.tsx
+++ b/packages/plugins/components/src/NavBar/navbar.stories.tsx
@@ -9,7 +9,7 @@ export default {
   component: NavBar,
 } as ComponentMeta<typeof NavBar>;
 
-const useStyles = makeStyles((theme: Theme) => ({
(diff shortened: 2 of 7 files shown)


ursucarina (approved):

packages/plugins/components/package.json
```diff
@@ -13,8 +13,8 @@
     },
     "scripts": {
         "build": "yarn build:esm && yarn build:cjs",
-        "build:esm": "tsc --module esnext --outDir lib/esm",
-        "build:cjs": "tsc",
+        "build:esm": "tsc --module esnext --outDir lib/esm -p ./tsconfig.build.json",
+        "build:cjs": "tsc -p ./tsconfig.build.json",
```
eugenejahn, line 17: nit: prefer use --project. easier to understand 

 --project, -p  Compile the project given the path to its configuration file, or to a folder with a 'tsconfig.json'.

packages/plugins/components/src/NavBar/navbar.stories.tsx
```diff
@@ -9,7 +9,7 @@ export default {
   component: NavBar,
 } as ComponentMeta<typeof NavBar>;
 
-const useStyles = makeStyles((theme: Theme) => ({
+const useStyles = makeStyles((_theme: Theme) => ({
```
eugenejahn, line 12: just wonder what is the difference? I saw in the code sometimes we use theme and sometimes use _theme

packages/zapp/console/tsconfig.json
```diff
@@ -12,5 +12,9 @@
         "noImplicitOverride": false
     },
     "references": [{ "path": "../../plugins/components" }],
-    "include": ["src/**/*"]
+    "include": [
+        "src/**/*",
+        // TODO: *.json could be removed when tsconfig.build.json would be properly consumed by webpack
+        "src/**/*.json"
```
eugenejahn, line 18: just wonder isn't 
`"src/**/*` include `"src/**/*.json`

eugenejahn (dismissed):
LGTM

packages/plugins/components/package.json
anrusina: sounds good, will update

packages/plugins/components/src/NavBar/navbar.stories.tsx
anrusina: '_' prefix let compiler now that even though this parameter is not in use at tis point, we want to keep it named in code. Otherwise eslint will complain about unused parameter.

None of used variables should have `_` prefix. If variable is not used we can either delete it or add _ prefix.
I decided to add prefix here, as I know that I will use it in my next PR :)

packages/zapp/console/tsconfig.json
anrusina: 🙃 I know it's sound funky, but no `**/*` doesn't include `*.json` files (and some other file types too). It was specifically done by TS - to avoid accidental addition of big files, which are rarely should be part of the project.
As webpack will currently include test/spec/ mock files, we would need to build in few jsons that our tests are relied upon. (for now)

eugenejahn (approved):

Issue #461, opened 2022-05-12
Add "rerun" ability for running tasks independently 


jsonporter:
To build a richer debugging experience, we want to provide users the ability to rerun a task independently of its parent workflow. To understand the use-case, Imagine a user has a large workflow with 12 nodes, that takes 10 hours to run, but the 11th node is failing. The fix will be likely be outside of console (eg, python) but testing the fix in our current flow would require the user to rerun the entire workflow (10 hours) to validate. However, in this new flow the we will render "rerun" buttons at various places in the execution details views that allow a user to run just that single failing task _with the same real-world inputs the previous nodes_ provided it. Once the user has fixed the bug, they can rerun the parent workflow

Acceptance Criteria:
 - Render rerun button in execution details slide-out panel
 - Render rerun icon (inline) on the execution table row
 - Clicking will open a launch form with the values that were used in the previous execution
 - Users can rerun with new values and these will not trigger the parent workflow.
 
Comps:
<img width="1082" alt="Screen Shot 2022-05-12 at 10 49 59 AM" src="https://user-images.githubusercontent.com/84735036/168137635-342ac296-4dba-406e-a16f-3fcd17c2e258.png">
<img width="1077" alt="Screen Shot 2022-05-12 at 10 49 47 AM" src="https://user-images.githubusercontent.com/84735036/168137642-56203b0d-8f69-4d80-8d80-0421f2698646.png">

PR #488, opened 2022-05-19
Eugene/rerun done

eugenejahn:
# TL;DR
Add the rerun feature in execution page 
- have a rerun icon 
-
image

- have rerun button in side panel 
image

- after click the button or the icon, will show the launch task popup

link to test: https://localhost.demo.nuclyde.io:3000/projects/flytesnacks/domains/development/executions/av4zwbjvp5g8fmtdcm4m?duration=all

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

https://github.com/flyteorg/flyteconsole/issues/461

## Follow-up issue
_NA_
OR


packages/zapp/console/src/components/Launch/LaunchForm/LaunchFormDialog.tsx
```diff
@@ -0,0 +1,38 @@
...
+import * as React from 'react';
+import { LaunchForm } from 'components/Launch/LaunchForm/LaunchForm';
+import { ResourceIdentifier, ResourceType } from 'models/Common/types';
+
+function getLaunchProps(id: ResourceIdentifier) {
+  if (id.resourceType === ResourceType.TASK) {
+    return { taskId: id };
+  }
+
+  return { workflowId: id };
```
anrusina, line 11: nit: in future the different resource type can be provided here, which will break things without a clear reason why.

In such cases it's better to either:

- Use switch/case statement with `default` return which includes debug.log() in case if unexpected resource type was provided

OR

- Use if statemets for all expected types, and regular return at the ens of the function for unexpected

anrusina (commented):
Please update your PR description

- title: similar to your commits - should include scope as `feat: description`, `fix:descrition`, etc. as master brunch releases are triggered depending on this words.
- description it self, to remove default strings like: TL;DR, Follow up issue section if it wasn;t modified by you, Complete Description section - if nothing there was added,
- properly mark checkboxes.

Will do final review, after values would be populated into the form.

Good work!

packages/zapp/console/src/components/Executions/Tables/NodeExecutionActions.tsx
```diff
@@ -0,0 +1,90 @@
...
+import { useNodeExecutionData } from 'components/hooks/useNodeExecution';
+import { TaskInitialLaunchParameters } from 'components/Launch/LaunchForm/types';
+import { literalsToLiteralValueMap } from 'components/Launch/LaunchForm/utils';
+import { NodeExecutionsTableState } from './types';
+import { useNodeExecutionContext } from '../contextProvider/NodeExecutionDetails';
+import { NodeExecutionDetails } from '../types';
+import t from './strings';
+
+export const NodeExecutionActions: React.FC<{
+  className?: string;
```
anrusina, line 17: Looks like className is never used, even if provided. can you please remove it from here?

packages/zapp/console/src/components/Launch/LaunchForm/LaunchFormDialog.tsx
```diff
@@ -0,0 +1,48 @@
...
+  if (id.resourceType === ResourceType.TASK) {
+    return { taskId: id };
+  } else if (id.resourceType === ResourceType.WORKFLOW) {
+    return { workflowId: id };
+  }
+  throw new Error('Unknown Resource Type');
+}
+
+export const LaunchFormDialog: React.FC<{
+  className?: string;
```
anrusina, line 20: same here, calssName is never used even if provided. We shouldn't add props just in case, until we know that the would be used.

packages/zapp/console/src/components/common/Icons/RerunIcon.tsx
```diff
@@ -0,0 +1,20 @@
+import * as React from 'react';
+import { IconProps } from './interface';
+
+export const RerunIcon: React.FC<IconProps> = ({ size = 18, className, onClick }) => {
```
anrusina, line 4: Please replace React.FC<> with a function as  `= () =>`.
React.FC has t's own downsides and planned to be deprecated. We should avoid using it.  Here is some explanation why: https://github.com/facebook/create-react-app/pull/8177
```
export const RerunIcon = (props: SVGProps): JSX.Element => (
<svg>...</svg>
)
```
If you feel like it `JSX.Element` return type can be replaced with more broader `React.ReactNode one`.

packages/zapp/console/src/components/common/Icons/RerunIcon.tsx
```diff
@@ -0,0 +1,20 @@
+import * as React from 'react';
+import { IconProps } from './interface';
```
anrusina, line 2: Sorry,  I asked you to put RerunIcon as Svg, but I forgot that we curently have 2 InfoIcons in the system.
To properly support plugin system in future we need to decouple our current "monolith" (highly intertwined) application into packages. In this system all icons should leave in one layer: Inside `flyteconsole/composites/ui-atoms/src/Icons`. In future all Icons we currently have would be shown in Storybook.

Can you please rename this file to index.tsx and  move it to `flyteconsole/composites/ui-atoms/src/Icons/RerunIcon`
Then you will need to add `export { InfoIcon } from './InfoIcon';` inside `src/Icons/index.tsx`
then import it as : `import { RerunIcon } from '@flyteconsole/ui-atoms';`

packages/zapp/console/src/components/common/Icons/RerunIcon.tsx
anrusina: Please replace it in all 4 places you were adding it and ensure to use props declaration instead of inline "name extraction" (not sure how it called in english).

I know it often seems redundant at first, but props is better coding style, simplifies future modification of props and allow easier inheritance of components or types in future. 
```
interface SomeProps {
 prop1: prop1type; ...
}
const Some = (props: SomeProps) => {...}
```

anrusina (commented):
Looked through. code, haven't performed smoke test. will leave it to Jason.

jsonporter:
Just updating the with offline conversations: "This should have input `s` as  `133`, right?
https://localhost.demo.nuclyde.io:3000/projects/flytesnacks/domains/development/executions/actdj9ws89vkr6xmkbs8?duration=all&fromExecutionNav=true"

packages/zapp/console/src/components/common/Icons/RerunIcon.tsx
eugenejahn: fixed

packages/zapp/console/src/components/common/Icons/RerunIcon.tsx
eugenejahn: fixed

packages/zapp/console/src/components/Launch/LaunchForm/LaunchFormDialog.tsx
eugenejahn: fixed

jsonporter (approved):
lgtm

PR #498, opened 2022-05-27
fix: Merge Devmain Back to Master #patch

anrusina:
This PR merges a set of previously done work into master, and should trigger a patch (0.0.X) release. 

Items are cherry-picked from devmain and would be merged as set of commits same way you see commits in this PR (to keep the history)

No additional changes were added.
Performed smoke test locally.


jsonporter (approved):
lgtm

Issue #406, opened 2022-04-18
[BUG] Console showing subworkflows as unknown


wild-endeavor:
## Describe the bug
When running a complex workflow with nested subworkflows and launch plans, Console is not correctly rendering the graph or the node types.

## To reproduce

The following command was run on
* flytekit version `3a57bd79bd6d0de8aaf336fb6477cfa41d2851af`
* flytesnacks version `fca5d5440c68cdc92589903ddfe6ae0ad8bda895` (branch `nested-updates`)

```
pyflyte run --remote core/control_flow/subworkflows.py:other_root_wf --a 15
```
https://demo.nuclyde.io/console/projects/flytesnacks/domains/development/executions/f83c218a9e72b40b7924?duration=all

For completeness, the `config.yaml` is
```
$ cat ~/.flyte/config.yaml
admin:
  endpoint: dns:///demo.nuclyde.io
  authType: Pkce
  insecure: false
logger:
  show-source: true
  level: 1

storage:
  type: stow
  stow:
    kind: s3
    config:
      auth_type: iam
      region: us-east-2
  container: "flyte-demo"
  cache:
    max_size_mbs: 512
    target_gc_percent: 70
  limits:
    maxDownloadMBs: 10
```

To avoid issues with the new `pyflyte run` command, I also used the old `pyflyte package` and `flytectl register` commands but the result was the same
```
~/go/src/github.com/flyteorg/flytesnacks/cookbook (nested-updates) $ pyflyte --pkgs core package --fast -i ghcr.io/flyteorg/flytekit:py3.8-latest

~/go/src/github.com/flyteorg/flytesnacks/cookbook (nested-updates) $ flytectl -p flytesnacks -d development -c ~/.flyte/demo-config.yaml register files --archive ~/go/src/github.com/flyteorg/flytesnacks/cookbook/flyte-package.tgz --version test_nested_updates_json1
```
https://demo.nuclyde.io/console/projects/flytesnacks/domains/development/executions/aklh492hn5xrj548bvcg?duration=all

### Remove the launch plan
If you however remove the launch plan and just call the subworkflow directly:
https://github.com/flyteorg/flytesnacks/compare/nested-updates...nested-updates-no-lp

It works:
https://demo.nuclyde.io/console/projects/flytesnacks/domains/development/executions/f884541a45092498f9b5?duration=all

### Are you sure this issue hasn't been raised already?

- [X] Yes

### Have you read the Code of Conduct?

- [X] Yes

PR #570, opened 2022-08-19
fix: console showing subworkflows as unknown

olga-union:
Signed-off-by: Olga Nad <olga@union.ai>

# TL;DR
Some subworkflows compared non-empty objects with `null`, which resulted in the failure to extract fields on a `null` and the whole stack of unresolved computational steps - adding an early exit out of the comparison function, if one of the parameters is null/undefined, resolves the issue with improper populated data in the node executions table.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Demo
Before: 
Screen Shot 2022-08-19 at 10 33 49 AM

After:
Screen Shot 2022-08-19 at 10 34 17 AM

## Tracking Issue
fixes https://github.com/flyteorg/flyteconsole/issues/406

## Follow-up issue
_NA_

### DIFF
--- a/packages/zapp/console/src/components/Executions/Tables/nodeExecutionColumns.tsx
+++ b/packages/zapp/console/src/components/Executions/Tables/nodeExecutionColumns.tsx
@@ -3,8 +3,7 @@ import { formatDateLocalTimezone, formatDateUTC, millisecondsToHMS } from 'commo
 import { timestampToDate } from 'common/utils';
 import { useCommonStyles } from 'components/common/styles';
 import { isEqual } from 'lodash';
-import { CatalogCacheStatus, NodeExecutionPhase } from 'models/Execution/enums';
-import { TaskNodeMetadata } from 'models/Execution/types';
+import { NodeExecutionPhase } from 'models/Execution/enums';
 import * as React from 'react';
 import { useNodeExecutionContext } from '../contextProvider/NodeExecutionDetails';
 import { ExecutionStatusBadge } from '../ExecutionStatusBadge';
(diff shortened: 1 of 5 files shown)


packages/zapp/console/src/components/Executions/Tables/nodeExecutionColumns.tsx
```diff
@@ -107,15 +106,6 @@ const DisplayType: React.FC<NodeExecutionCellRendererData> = ({ execution }) =>
...
 
-const hiddenCacheStatuses = [CatalogCacheStatus.CACHE_MISS, CatalogCacheStatus.CACHE_DISABLED];
-function hasCacheStatus(taskNodeMetadata?: TaskNodeMetadata): taskNodeMetadata is TaskNodeMetadata {
-  if (!taskNodeMetadata) {
-    return false;
-  }
-  const { cacheStatus } = taskNodeMetadata;
-  return !hiddenCacheStatuses.includes(cacheStatus);
-}
-
```
olga-union, line 118: removed unused code

packages/zapp/console/src/components/WorkflowGraph/utils.ts
```diff
@@ -35,6 +35,11 @@ export function isExpanded(node: any) {
  * @returns     boolean
  */
 export const checkIfObjectsAreSame = (a, b) => {
+  // if one of the objects is null (undefined), objects can't be the same
+  if ((!a || !b) && a != b) {
```
ursucarina, line 39: nit: if these are objects, then the equality check might not be enough.

How deep does the comparison need to be? would `lodash.isEqual(a, b) `be a better alternative?

ursucarina (approved):

packages/zapp/console/src/components/WorkflowGraph/utils.ts
olga-union: That's a good idea - we can completely replace this helper with `_.isEqual`, I was working on top of existing code and thought about using `_.has` which doesn't help with `null` values, but replacing the whole this with isEqual is even better!

ursucarina (approved):

jsonporter (approved):
lgtm

Issue #571, opened 2022-08-25
Bug: Launch form with 'Optional' I/O displays error saying task cannot be run in console


jsonporter:
User reported in Slack channel that a workflow with optional inputs/outputs did not work. Flyteconsole currently should support these data types; investigate the issue and fix if necessary. 

Context: https://unionai.slack.com/archives/C02DG114DJN/p1661295278179679
<img width="421" alt="image (10)" src="https://user-images.githubusercontent.com/84735036/186715976-d9cec2e9-8994-4c2d-886f-5982317ef2ba.png">

jsonporter:
fixes merged

PR #577, opened 2022-09-02
added none type in union type

eugenejahn:

since we don't support none type, there is an bug in Union Type when using optional type. The reason is optional type will use none type. Since we added the none type support in flyteconsole, there shouldn't have any bugs on union type. 

Issue: https://github.com/flyteorg/flyteconsole/issues/571


jsonporter (approved):
lgtm

PR #629, opened 2022-10-27
fix: failing/empty graph and timeline tabs upon relaunch data updates

olga-union:
# TL;DR
Re-renders were throwing errors/displaying empty pages, when components got empty data blips while waiting on workflow status updates, from which they couldn't recover. Updated effect in Timeline, and added a skip of graph render in case of empty `dag` variable (displaying a message that the graph is empty), as the graph state is complex and not easily adjustable from one point from the empty state, plus saving on render/computation cost in this scenario.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Demo
Graph 
https://user-images.githubusercontent.com/101579322/198358307-37170ee6-a2cc-494a-bf32-b9c21503de87.mov

Timeline
https://user-images.githubusercontent.com/101579322/198358362-6a90ad50-3c41-4202-96ac-3dfe20ea18ac.mov

## Tracking Issue
_NA_

## Follow-up issue
_NA_


packages/zapp/console/src/components/flytegraph/ReactFlow/transformDAGToReactFlowV2.tsx
```diff
@@ -236,6 +236,11 @@ export const buildGraphMapping = (props): ReactFlowGraphMapping => {
   const parse = (props: ParseProps) => {
     const { contextNode, contextParent, rootParentNode, nodeDataProps } = props;
     let context: ReactFlowGraph | null = null;
+    if (!contextNode) {
```
olga-union, line 239: we should no longer reach this path with empty `contextNode`, but added a debugger message, just in case, for easier investigation of potential bugs in future.

jsonporter (approved):

PR #640, opened 2022-11-28
Remove Google Analytics Key

rafaelraposospot:
# TL;DR

Replaces a Google Analytics key with an empty value.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

fixes https://github.com/flyteorg/flyteconsole/issues/641

## Follow-up issue
_NA_

### DIFF
--- a/packages/zapp/console/env.js
+++ b/packages/zapp/console/env.js
@@ -37,7 +37,7 @@ const STATUS_URL = process.env.STATUS_URL;
 
 // Configure Google Analytics
 const ENABLE_GA = process.env.ENABLE_GA || false;
-const GA_TRACKING_ID = process.env.GA_TRACKING_ID || 'G-0QW4DJWJ20';
+const GA_TRACKING_ID = process.env.GA_TRACKING_ID || '';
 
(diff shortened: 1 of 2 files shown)


packages/zapp/console/env.js
```diff
@@ -37,7 +37,7 @@ const STATUS_URL = process.env.STATUS_URL;
 
 // Configure Google Analytics
 const ENABLE_GA = process.env.ENABLE_GA || false;
-const GA_TRACKING_ID = process.env.GA_TRACKING_ID || 'G-0QW4DJWJ20';
```
honnix, line 40: Maybe we should just disable GA by default, instead of configuring tracking id to be invaild.

honnix: Looks like it is already default to `false`, so maybe this is just a documentation problem after all?

rafaelraposospot: It would still make sense to remove the Key even for cleaness or even security concerns if it's a valid GA key.

jsonporter (approved):

PR #676, opened 2023-01-30
fix: execution view children fetch on demand refactor

olga-union:
Signed-off-by: Olga Nad <olga@union.ai>

This fixes a previous refactor on node details UX by placing execution data into a single shared context between List, Graph and Timeline views (to avoid each view re-fetching already fetched data). However, the previous solution would front load the data b recursively fetching each execution's children which we found caused major issues for larger dynamic workflows.

To fix this issue this PR will implement an on-demand fetch for dynamic workflows.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin
 - [ ] Chore

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description

- Removed childGroups fetch (recursive) from `ExecutionNodeViews` page
- Moved childGroups fetch to `onToggle` in Table and Timeline components - requests trigger when user clicked on expander
- Moved map tasks fetch to details page to back fill one-off nodeExecutions data, and to `ReactFlowGraphComponent` to batch them for all known nodeExecutions, so the graph is rendered properly (this will slow down the graph tab, but based on user's feedback, it's ok, if graph tab takes some time to load).
- Updated Graph nesting logic to support unexpanded subworkflows at level 0

## Known issues
- [ ] some unit tests are broken
- [x] add childGroups fetch in graph on node click
- [x] fix rerender of graph on childGroups fetch
- [x] fix `compiledWorkflowClosure` to check for `is_dynamic` field (available in node execution's metadata) instead of pulling multiple requests on repeat
- [ ] test filters in Table tab
- [x] display parent node as `nested` in graph to indicate it has children
- [ ] child task type stays in Loading state in the Details Sidebar

## Follow-up issue
_NA_

## Test workflows
- normal `core.control_flow.subworkflows.parent_wf`
- dynamic `core.control_flow.dynamics.wf`
- nested parent `core.control_flow.subworkflows.nested_parent_wf`
- map task `flyte.workflows.example_map_task.wf`
- small fan-out dynamic `workflows.branching_dynamic_example.wf`

### DIFF
--- a/packages/console/src/components/Executions/ExecutionDetails/ExecutionNodeViews.tsx
+++ b/packages/console/src/components/Executions/ExecutionDetails/ExecutionNodeViews.tsx
@@ -1,20 +1,12 @@
-import * as React from 'react';
+import React, { useEffect } from 'react';
 import { Tab, Tabs } from '@material-ui/core';
 import { makeStyles, Theme } from '@material-ui/core/styles';
 import { WaitForQuery } from 'components/common/WaitForQuery';
 import { DataError } from 'components/Errors/DataError';
 import { useTabState } from 'components/hooks/useTabState';
 import { secondaryBackgroundColor } from 'components/Theme/constants';
-import {
-  Execution,
-  ExternalResource,
-  LogsByPhase,
-  NodeExecution,
-} from 'models/Execution/types';
-import { useEffect, useMemo, useState } from 'react';
+import { Execution } from 'models/Execution/types';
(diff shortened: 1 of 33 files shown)


james-union (dismissed):
Very good job! LGTM.

ursucarina (dismissed):

ursucarina (approved):

PR #680, opened 2023-02-07
Perf Refactor: Some parent nodes don't load children

james-union:
https://github.com/unionai/cloud/issues/1645

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR

### DIFF
--- a/packages/console/src/components/common/utils.ts
+++ b/packages/console/src/components/common/utils.ts
@@ -27,13 +27,15 @@ export function measureText(fontDefinition: string, text: string) {
  */
 export const checkForDynamicExecutions = (allExecutions, staticExecutions) => {
   const parentsToFetch = {};
+  const executionsByNodeId = {};
   for (const executionId in allExecutions) {
+    const execution = allExecutions[executionId];
+    executionsByNodeId[execution.id.nodeId] = execution;
     if (!staticExecutions[executionId]) {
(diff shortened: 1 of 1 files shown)


packages/console/src/components/common/utils.ts
```diff
@@ -27,13 +27,15 @@ export function measureText(fontDefinition: string, text: string) {
  */
 export const checkForDynamicExecutions = (allExecutions, staticExecutions) => {
   const parentsToFetch = {};
+  const executionsByNodeId = {};
   for (const executionId in allExecutions) {
+    const execution = allExecutions[executionId];
```
olga-union, line 32: let's keep the `dynamicExecution` name, so we don't trigger the changes here, and `executionsByNodeId` can be transformed to `dynamicExecutionsByNodeId`

jsonporter (approved):

packages/console/src/components/common/utils.ts
james-union: @olga-union. As this execution can be dynamic or static, so I don't think we can name it as dynamicExecution.

PR #693, opened 2023-03-01
fix: type error on invalid nested json input

4nalog:
# TL;DR

This PR fixes the `parseJson` utility function that maps inputs to a JSON form to handle situations when a nested object is optional or has value of null.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description

- fix: `parseJson` helper function does not handle `null` values

## Tracking Issue

fixes https://github.com/flyteorg/flyte/issues/3157

## Follow-up issue
_NA_


packages/console/src/components/Launch/LaunchForm/StructInput.tsx
```diff
@@ -69,7 +69,10 @@ export const StructInput: React.FC<InputProps> = props => {
       literalType?.metadata?.fields?.definitions?.structValue?.fields,
     );
 
-    if (keys[0]) {
+    if (keys.length > 1) {
+      // If there are multiple keys, we can't render a form because of not supporting nested structs
+      jsonFormRenderable = false;
```
jsonporter, line 74: I think we want to render a free-form text area in this case; they can paste in JSON?

jsonporter (approved):

Issue #644, opened 2022-12-05
Mapped Tasks not showing cache status correctly.


jsonporter:
Users are noticing some inconsistent behavior in regards to the cache status for mapped tasks. It seems like cache values are only out of sync in running/unknown phases (and eventually rectify once the parent task is `succeeded`). This ticket will track root causing and fixing the issue.

PR #711, opened 2023-03-07
Mapped Tasks not showing cache status correctly.

james-union:
https://github.com/flyteorg/flyteconsole/issues/644

Two issues
(1) It renders cache status incorrectly because it depended on the index of the logs instead of the id. While running, the logs are not in the original order (they are in order of queued/success instead of original order). So need to find correct cache status based on id
(2) the useStyles/hooks should not be used after component is rendered. It broke the app when the execution status is updated. I managed to place it before any render

To reproduce

For this workflow,
https://development.uniondemo.run/console/projects/flytesnacks/domains/development/workflows/map-task.my_cached_map_workflow_1?duration=all

- Click `Launch Workflow` button
- For the parameter, we need to mix some cached numbers and non-cached numbers to reproduce this issue. Right now, 0 ~ 8 are cached. (This will be changed if someone run this more). So for example, if 0 ~ 8 are cached, we can set the param as `[9,1,0,12,11,3]`
- Then while running the mapped task, if you go to the Map Execution panel on the right side (by navigating to the execution details and and click the first Map task), you will see some queued & succeeded tasks. Actually the problem is the icons next to the task names. If those are already queued, then it should be the refresh icon which identifies already cached.
- To reproduce the second issue, you should wait until the Map task is completed while open this right panel. Once it is updated, the right panel will be broken.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

fixes https://github.com/flyteorg/flyte/issues/644


ursucarina:
Did you link the correct issue?

james-union:
> Did you link the correct issue?

Yes. This is for the frontend part.

packages/console/src/components/common/MapTaskExecutionsList/TaskNameList.tsx
```diff
@@ -68,7 +72,7 @@ export const TaskNameList = ({
               color={log.uri ? 'primary' : 'textPrimary'}
               onClick={log.uri ? handleClick : undefined}
               key={taskLogName}
```
ursucarina, line 74: can you add this key to the div that's wrapping this please?
 <div
            style={{
              display: 'flex',
              alignItems: 'center',
            }}
            key={taskLogName}
          >

ursucarina (changes requested):

PR #738, opened 2023-03-31
feat: show launchplan in execution table

pradithya:
# TL;DR
Display Launch Plan name in the workflow execution table and link it to the launch plan details page.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested

## Complete description

Display Launch Plan name in the workflow execution table and link it to the launch plan details page.

Screenshot 2023-04-01 at 9 56 39 AM

https://user-images.githubusercontent.com/4023015/229260444-fc10c67c-ff69-4076-9db4-38b99c7f0bdb.mov

## Tracking Issue

[https://github.com/flyteorg/flyte/issues/3555](https://github.com/flyteorg/flyte/issues/3555)

## Follow-up issue
_NA_


jsonporter:
Hey thanks for this PR; looking through the code this looks largely okay my only concern is around how we want to handle archived launch plans (in general) so I've forwarded this PR to our design/UX team and we'll try to get some feedback in the next day or two 👍

kumare3:
@jsonporter I actually think it makes sense to show launchplan, as you trigger launchplans and they become executions. For a single workflow, there could be multiple launch plans. from utlity pov, i like this PR +1

packages/console/src/components/Executions/Tables/WorkflowExecutionTable/styles.ts
```diff
@@ -10,6 +10,10 @@ export const useStyles = makeStyles((theme: Theme) => ({
     flexBasis: workflowExecutionsTableColumnWidths.name,
     whiteSpace: 'normal',
   },
+  columnLaunchPlan: {
```
jsonporter, line 13: Wondering how this would handle a really large string; could we add something like `overflow:hidden` here?

hongxunjin:
In the previous Figma design proposal.(shown down below) , the access to the launchplan was placed at the same level as "archive" and "view I/O" , as a "hovering over icon". However, the question remains: how prominent should the launchplan's access be?

If our goal is to display the launchplan's relevant information (such as name and version) upfront, then the proposal in this PR seems reasonable.

image

pradithya:
> In the previous Figma design proposal.(shown down below) , the access to the launchplan was placed at the same level as "archive" and "view I/O" , as a "hovering over icon". However, the question remains: how prominent should the launchplan's access be?
> 
> If our goal is to display the launchplan's relevant information (such as name and version) upfront, then the proposal in this PR seems reasonable.
> 
> image

The pain point that I am addressing is the need for additional clicks to figure out which launch plan is used when executing a workflow. Currently, it's possible to figure that out by opening each execution page (with a few extra clicks). But it's becoming painful when there are multiple launch plans executing the same workflow. Having this as a column greatly helps to minimize friction. 
The issue https://github.com/flyteorg/flyte/issues/3555 mentioned filtering capability by launch plan which is also useful to further improve the UX.

tsheiner:
This PR seems like an improvement to me. 

In general, bringing information up, not hiding it behind clicks, is a good thing. It becomes a negative only when the additional info is distracting, not related to primary task of the view, or it makes the view too complex and hard to read.

Neither of those concerns appear to apply in this case.

@jsonporter can you say more about this: "my only concern is around how we want to handle archived launch plans (in general)..."

jsonporter:
Yeah just that "archiving" a launch plan is a little odd in that it doesn't actually prevent executions (ie, you can launch a WF from an archived launch plan). That said, we discussed the edge cases and they seem fine; i'm good with this PR if everyone else is 👍

jsonporter (approved):

PR #765, opened 2023-05-19
fix: dynamic-node-tasks

4nalog:
This PR fixes a bug when viewing children of dynamic task where the various information like the task type/template was not correctly being shown. 

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
The issue was caused by incorrectly referencing task definitions from the (root) `compiledWorkflowClosure` rather than dynamic execution's `workflowClosure`. The result was that row values like task type as well as information in the execution detail panel (slide-out) were null and showing empty values.

fixes [#3455 ](https://github.com/flyteorg/flyte/issues/3455)

## Follow-up issue
_NA_

### DIFF
--- a/packages/console/package.json
+++ b/packages/console/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@flyteorg/console",
-  "version": "0.0.36",
+  "version": "0.0.37",
   "description": "Flyteconsole main app module",
   "main": "./dist/index.js",
   "module": "./lib/index.js",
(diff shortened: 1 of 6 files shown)


4nalog:
This is working well:

Verified with the workflow:
https://localhost.development.uniondemo.run:3000/console/projects/flytesnacks/domains/development/executions/admspfjxlzcl6q55hcfc?duration=all

All fields are populated correctly from the task type and the descriptions are also correct for the task template on the slideout

ursucarina (dismissed):

jsonporter (approved):

PR #809, opened 2023-08-01
fix: invalid item in dropdown

etaldot:
## _Read then delete this section_

_- Make sure to use a concise title for the pull-request._

_- Use #patch, #minor or #major in the pull-request title to bump the corresponding version. Otherwise, the patch version
will be bumped. [More details](https://github.com/marketplace/actions/github-tag-bump)_

# TL;DR
_Please replace this text with a description of what this PR accomplishes._

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR

### DIFF
--- a/packages/console/src/components/Breadcrumbs/components/BreadcrumbPopover.tsx
+++ b/packages/console/src/components/Breadcrumbs/components/BreadcrumbPopover.tsx
@@ -1,5 +1,5 @@
 import React, { useMemo } from 'react';
-import { useQuery } from 'react-query';
+import { useQuery, useQueryClient } from 'react-query';
 import isEmpty from 'lodash/isEmpty';
 import {
   Box,
@@ -24,6 +24,7 @@ import { BreadcrumbEntity, BreadcrumbPopoverInterface } from '../types';
 
 const BreadcrumbPopOver = (props: BreadcrumbPopoverInterface) => {
   const history = useHistory();
+  const queryClient = useQueryClient();
 
   const {
     isLoading,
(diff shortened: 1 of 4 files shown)


FrankFlitton (approved):

packages/console/src/components/Breadcrumbs/components/BreadcrumbPopover.tsx
```diff
@@ -24,16 +24,17 @@ import { BreadcrumbEntity, BreadcrumbPopoverInterface } from '../types';
...
   const history = useHistory();
+  const queryClient = useQueryClient();
 
   const {
     isLoading,
     error,
     data: popoverQueryData,
   } = useQuery(
-    `breadcrumb-list-${props.id}`,
+    `breadcrumb-list-${props.id}-${props.value}`,
```
ursucarina, line 34: I don't think we should fetch the project list every time the domain changes. I think we should separate the api call and the url generation step so we don't make unnecessary network calls

ursucarina (changes requested):

packages/console/src/components/Breadcrumbs/components/BreadcrumbPopover.tsx
FrankFlitton: This would be a good enhancement, we'd probably need a query hash key, the formatter and data as separate params in the Breadcrumb objects. Some breadcrumbs don't change from url to url etc... 

A note here is that it only fetches once the user opens that dropdown.

ursucarina: This is how the original code worked, not an enhancement. If we separate the list request from the static url generation this wouldn't be a problem

ursucarina (changes requested):

ursucarina (dismissed):

ursucarina (approved):

ursucarina (dismissed):

ursucarina (approved):

FrankFlitton (approved):

jsonporter (approved):

PR #824, opened 2023-10-19
feat: show io urls & flyteremote usage

ursucarina:
# TL;DR
* Shows Context Panel IO URLs
* Shows Context Panel IO FlyteRemote usage
* Shows Context Panel node execution logs

|        |           |
| ------------- |:-------------:| 
| image| image |
| image | image |

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR

### DIFF
--- a/package.json
+++ b/package.json
@@ -106,6 +106,7 @@
     "prettier": "^2.8.3",
     "react": "^16.13.1",
     "react-dom": "^16.13.1",
+    "react-syntax-highlighter": "^15.5.0",
     "semantic-release": "^21.0.7",
     "serve-static": "^1.12.3",
     "source-map-loader": "^4.0.1",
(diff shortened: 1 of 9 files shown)


jpvotta:
@pingsutw how does this work with maptask (subtasks)? I believe we are doing some parsing in the FE in order to allocate the input/output list elements into inputs/outputs of each subtask respectively. I am guessing that we would want to include these new visual elements in the maptask parent (since the literalmap will include the full list of inputs/outputs including all subtasks), but not for each subtask (since this would be identical to the info in the parent).

pingsutw:
> I believe we are doing some parsing in the FE in order to allocate the input/output list elements into inputs/outputs of each subtask respectively.

Right, so there is no flyte url for sub task. we can just show it only in the parent for now.
btw, UI looks pretty good now, thank you so much.

one nit: we should not show the copy URL and code snippet if the input or output is empty.

pingsutw:
cc @wild-endeavor

FrankFlitton (approved):

PR #861, opened 2024-04-04
fix: Broken breadcrumbs when using custom BASE_URL

ddl-rliu:
# TL;DR
Typo – causes breadcrumbs to not appear in flyteconsole. To fix, use BASE_URL not BASE_PATH.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

Missing breadcrumbs for custom BASE_URL like `/custom-base/console`

image

Expectation: breadcrumbs are present for custom BASE_URL Flyte instance.

image

This happens because when `BASE_PATH` is used instead of `BASE_URL`, the logic that generates the breadcrumbs at `breadcrumbRegistry.breadcrumbBuilder` ends up using the wrong `BASE_URL` value (this starts from `makeUrlSegments`). This causes it to parse the window location incorrectly, and the resulting generated breadcrumbs is missing the expected entries. 

## Tracking Issue

## Follow-up issue
_NA_
OR

### DIFF
--- a/packages/oss-console/src/components/Breadcrumbs/registry/index.ts
+++ b/packages/oss-console/src/components/Breadcrumbs/registry/index.ts
@@ -106,7 +106,7 @@ export class BreadcrumbRegistry {
 
   static makeUrlSegments(location: Location, projectId = '', domainId = '') {
     const pathName = location.pathname;
-    const basePath = process.env.BASE_PATH || '/console';
+    const basePath = process.env.BASE_URL || '/console';
 
     // Remove first occurence of base path
     const pathNameWithoutBasePath = pathName.replace(basePath, '');


packages/oss-console/src/components/Breadcrumbs/registry/index.ts
```diff
@@ -106,7 +106,7 @@ export class BreadcrumbRegistry {
 
   static makeUrlSegments(location: Location, projectId = '', domainId = '') {
     const pathName = location.pathname;
-    const basePath = process.env.BASE_PATH || '/console';
```
ddl-rliu, line 109: I believe BASE_PATH is a typo here, since the `process.env` object looks like
```
{
    NODE_ENV: "production",
    ADMIN_API: "//localhost:30080",
    ADMIN_API_URL: "localhost:30080",
    BASE_URL: "/console",
    BASE_HREF: "",
    DISABLE_CONSOLE_ROUTE_PREFIX: "",
    MAINTENANCE_MODE: ""
}
```
(in the case of a custom path, it looks like `... BASEURL: "/custom-path/console"`

And double checking the repo code, BASE_PATH has no other instance of usage, pointing to this being a potential typo.

eapolinario:
@jsonporter , can you take a look?

eapolinario (approved):

PR #882, opened 2024-07-09
feat: display labels on execution detail page

blaketastic2:
# TL;DR
This PR adds the spec labels to the execution details page

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [X] Unit tests added

image

### DIFF
--- a/packages/oss-console/src/components/Executions/ExecutionDetails/ExecutionLabels.tsx
+++ b/packages/oss-console/src/components/Executions/ExecutionDetails/ExecutionLabels.tsx
@@ -1,0 +1,36 @@
+import React from 'react';
+import Chip from '@mui/material/Chip';
+import makeStyles from '@mui/styles/makeStyles';
+
+type ValuesType = {[p: string]: string};
+interface Props {
+  values: ValuesType;
+}
+
+const useStyles = makeStyles({
+  chipContainer: {
+    display: 'flex',
+    flexWrap: 'wrap',
+    width: '100%',
+    maxWidth: '420px'
+  },
+  chip: {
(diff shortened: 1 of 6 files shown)


JamesHoover:
This is great, thank you. A few items I would like to see:

1. Updated color, yellow is too prominent, update to a secondary color like grey.

2. Overflow states:
- what happens when a single label is very long?
- what happens when you have 3+ rows of labels?

3. Empty state, what happens when there are no labels?

4. Interactions, are these intended to be clickable or copyable?

blaketastic2:
Here's what it looks like with more(both longer and more rows)

blaketastic2:
Here's an updated version with the `info` color set on the Chip.

blaketastic2:
Empty state, the column doesn't show. Would you prefer the `---` behaviour instead?

Also, there is no functionality on the Chips. If someone wants to add something later(like copy to clipboard), I think it should be trivial. Do you think this is ok?

JamesHoover:
> Empty state, the column doesn't show. Would you prefer the `---` behaviour instead?
> 
> Also, there is no functionality on the Chips. If someone wants to add something later(like copy to clipboard), I think it should be trivial. Do you think this is ok?

Yes, I would prefer the `---` behavior for the empty state for consistency and discovery. Otherwise this looks good to me. Thank you for your contribution!

blaketastic2:
Ok, done.

image

jsonporter (approved):

PR #896, opened 2026-02-09
Feat: Allow setting advanced options on task launch like workflow launch

fg91:
# TL;DR
Currently one can set so-called "advanced options" like labels, annotations, raw output prefix, ... when launching a workflow. One cannot currently do so when launching a task.

This PR adds the same "advanced options" to the task launch menu (minus max parallelism which doesn't make sense for launching tasks).

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added
 - [x] Any pending items have an associated Issue

### DIFF
--- a/packages/oss-console/src/components/Launch/LaunchForm/LaunchTaskForm.tsx
+++ b/packages/oss-console/src/components/Launch/LaunchForm/LaunchTaskForm.tsx
@@ -1,4 +1,9 @@
+import Accordion from '@mui/material/Accordion';
+import AccordionDetails from '@mui/material/AccordionDetails';
+import AccordionSummary from '@mui/material/AccordionSummary';
 import DialogContent from '@mui/material/DialogContent';
+import Typography from '@mui/material/Typography';
+import ExpandMoreIcon from '@mui/icons-material/ExpandMore';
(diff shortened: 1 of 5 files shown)


packages/oss-console/src/components/Launch/LaunchForm/LaunchTaskForm.tsx
```diff
@@ -65,28 +72,46 @@ export const LaunchTaskForm: React.FC<LaunchTaskFormProps> = (props) => {
...
+                initialValue={state.context.defaultAuthRole}
+                ref={roleInputRef}
+                showErrors={state.context.showErrors}
+              />
+            ) : null}
+            {isEnterInputsState(baseState) ? (
+              <LaunchFormAdvancedInputs ref={advancedOptionsRef} state={state} />
+            ) : null}
+            <LaunchInterruptibleInput
+              initialValue={state.context.interruptible}
```
fg91, line 106: Moved into the "advanced options" menu as is the case for "Launch Workflow".

